PR #1, opened 2020-06-17
Feature/wjamieso/mapl refactor

WilliamJamieson:
Added in Aerosols_GridComp. Adjusted directory structure.


CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
+cmake_minimum_required (VERSION 3.17)
+cmake_policy (SET CMP0053 NEW)
+cmake_policy (SET CMP0054 NEW)
+
+project (
+        GOCART
+        VERSION 10.12.5
```
tclune, line 7: Start with 1.0.0.   The version from the GCM GridComp will rapidly diverge.   No really good option, but this is the convention we have used with other repos that split off.

And to be fair, you should probably call it 0.9.0 for now while we are getting things in place.

ESMF/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+add_subdirectory(MAPL)
```
tclune, line 1: Should this be "MAPL@" instead of "MAPL"

CMakeLists.txt
tclune: ...

ESMF/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
...
+#add_subdirectory(NUOPC)
```
tclune, line 2: Add a newline to end

ESMF/MAPL/Aerosol_GridComp/AeroCinderella_GridComp/AeroCinderella_StateSpecs.rc
```diff
@@ -0,0 +1,106 @@
+schema_version: 2.0.0
+component: DU
```
tclune, line 2: Wrong name.

ESMF/MAPL/Aerosol_GridComp/AeroCinderella_GridComp/Cinderella/FieldDictMap.F90
```diff
@@ -0,0 +1,10 @@
+module FieldDictMap_mod
```
tclune, line 1: Please use the convention we have in the rest of MAPL.  "Dict" is Python.  gFTL is based on STL, where these are Maps.

ESMF/CMakeLists.txt
tclune: Even more confused - I now found the yaml file and it has "@MAPL".  So this should not even work as it stands.

CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+# Set the default build type to release
+if (NOT CMAKE_BUILD_TYPE)
+    message (STATUS "Setting build type to 'Release' as none was specified.")
+    set (CMAKE_BUILD_TYPE Release CACHE STRING "Choose the type of build." FORCE)
+    # Set the possible values of build type for cmake-gui
+    set_property (CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS
+            "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
+endif ()
+
+set (DOING_GEOS5 YES)
```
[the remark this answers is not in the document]
mathomp4: My guess is this might be moot. I think it's only used in @GEOSchem_GridComp/Shared/Chem_Shared:
```cmake
if (DOING_GEOS5)
  list (APPEND srcs mod_diag.F90 diaglist.F90)
  list (APPEND srcs Chem_UtilMod.F90 Chem_StateMod.F90)
else ()
  list (APPEND srcs Chem_UtilMod.F90)
  list (APPEND srcs Chem_UtilMod_geos4.F90)
endif ()
```

It's there because it was there in GNU make/CVS land. If @mmanyin or @adarmenov say it's no longer needed, we can nuke that code all together.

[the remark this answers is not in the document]
mmanyin: I have scanned the GCM and CTM code (*F90) for Chem_UtilMPread; it is not used, so the geos4 code appears superfluous.  If @adarmenov agrees, we could just change the section above to
```
  list (APPEND srcs mod_diag.F90 diaglist.F90)
  list (APPEND srcs Chem_UtilMod.F90 Chem_StateMod.F90)
```
without checking  DOING_GEOS5.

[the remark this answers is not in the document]
adarmenov: Agreed.

tclune:
Note the legacy GOCART in this repo will never be used in GEOS.  GEOS will switch to this repo when GOCART_2G is complete.

CMakeLists.txt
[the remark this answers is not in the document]
tclune: Note - that the main copy in GEOSchem_GridComp should get this fix.

CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+
+list (APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/@cmake")
+include (esma)
+
+ecbuild_declare_project()
+
+# Generic DFLAGS
+# These should be relocated and/or eliminated.
+add_definitions(-Dsys${CMAKE_SYSTEM_NAME} -DESMA64)
+add_definitions(${MPI_Fortran_COMPILE_FLAGS})
```
tclune, line 38: This line should be removed.   Components that depend on MPI can just use
 ```
target_link_libraries(...   MPI::MPI_Fortran)
```

ESMF/CMakeLists.txt
tclune: Tom bad.   Did not realize that this was a new file when I was reading down.

PR #6, opened 2020-09-16
Add CircleCI

mathomp4:
This is to add CircleCI build testing with GOCART.


mathomp4:
Welp, need a newer CMake in the image! I knew it was coming...

PR #7, opened 2020-09-16
Remove mpeu dependency from GOCART2G

WilliamJamieson:
Small changes to GOCART2G, to eliminate the dependency on the `m_StrTemplate` provided by  `GMAO_Shared/GMAO_mpeu` in favor of the replacement in `MAPL_StringTemplate` provided by MAPL.

Note that I have not tested this yet, so I ask @gmao-esherman to please test this change before this is accepted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Changed
+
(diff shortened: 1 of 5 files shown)


tclune (changes requested):
Because Arlindo may see this layer, we need to follow his direction to not change the interface.  

Create a wrapper procedure StrTemplate that has the mpeu signature but that calls the MAPL `fill_grads_template` procedure.   Then do a PR to add that interface to MAPL.  When the MAPL change is released, a second change in GOCART can eliminate the redundant wrapper.

PR #10, opened 2020-09-24
Fixes optional rc bugs

### LINKED ISSUE #9: RC codes in Process_Library
There are many optional return codes (`rc`) in the Process_Library that are directly assigned to without checking if the `rc` is present. If one calls a routine without specifying the optional `rc` this will cause a runtime error (Segmentation fault).

WilliamJamieson:
The process library `GOCART2G_Process.F90` had several instances of an optional `rc` being set without first checking if it is present. This fixes those instances which resolves #9.

### DIFF
--- a/Process_Library/GOCART2G_Process.F90
+++ b/Process_Library/GOCART2G_Process.F90
@@ -198,7 +198,7 @@ CONTAINS
       emissions = emissions + (Ch_DU * du_src * emissions_)
     end do ! n
  
-   status = 0
+   status = __SUCCESS__
 
    __RETURN__(status)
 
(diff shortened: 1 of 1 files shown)


mathomp4:
Note: The CI failure is because the CircleCI config file is only on `develop` and not on `main` (or this branch) yet.

gmao-esherman:
Can we use the ___EXAMPLE__ commands? I thought this was MAPL related.

WilliamJamieson:
@gmao-esherman, you can use any of the macros defined at the top of the file (lines 1-7). They happen to imitate the MAPL ones.

tclune (changes requested):
I think these changes are not necessary.   The process library does not use optional output for rc.

tclune:
This PR requires a discussion.

WilliamJamieson:
Closing due offline discussion between @tclune and @amdasilva

PR #11, opened 2020-09-25
Fixes rc optional bugs

### LINKED ISSUE #9: RC codes in Process_Library

WilliamJamieson:
Fixes #9, without changing any `rc` interfaces.

### DIFF
--- a/Process_Library/GOCART2G_Process.F90
+++ b/Process_Library/GOCART2G_Process.F90
@@ -254,8 +254,8 @@ CONTAINS
 
         emissions_point(i,j,:) =  point_column_emissions
         end do
-   rc = 0
 
+      __RETURN__(__SUCCESS__)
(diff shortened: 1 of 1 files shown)


tclune (approved):

tclune:
I suspect Arlindo will complain when he sees the variant spellings of these macros.   But I'd leave them as is for now.   He'll have similar problems with my respelling elsewhere.    But I do want to review once again exactly that the standard (C) says about CPP tokens reserved for the system.  Technically they do not impact Fortran, but safest to align with that.

gmao-esherman (approved):
Model builds and runs with the hotfix.

PR #13, opened 2020-09-29
Merge rc bugfix into develop

WilliamJamieson:
Merge bug fixes from main to develop


WilliamJamieson:
Note that CircleCI will have failures when dealing with `main` until we merge the CircleCI configuration into main. I have tested that this does in fact build, both on my local machine and on discover.

PR #14, opened 2020-10-13
Fixed gfortran build errors

WilliamJamieson:
Fixed a few bugs that caused build failures when building with the gfortran compiler. All were related, namely logical variables cannot be compared with `==` and `\=`, they must instead be compared with `.eqv.` and `.neqv.` respectively.

### DIFF
--- a/ESMF/GOCART2G_GridComp/CA2G_GridComp/CA2G_GridCompMod.F90
+++ b/ESMF/GOCART2G_GridComp/CA2G_GridComp/CA2G_GridCompMod.F90
@@ -169,7 +169,7 @@ contains
 !   ------------------------
     call MAPL_GridCompSetEntryPoint (GC, ESMF_Method_Initialize,  Initialize, __RC__)
     call MAPL_GridCompSetEntryPoint (GC, ESMF_Method_Run, Run, __RC__)
-    if (data_driven /= .true.) then
+    if (data_driven .neqv. .true.) then
        call MAPL_GridCompSetEntryPoint (GC, ESMF_Method_Run, Run2, __RC__)
     end if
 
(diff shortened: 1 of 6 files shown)


ESMF/GOCART2G_GridComp/CA2G_GridComp/CA2G_GridCompMod.F90
```diff
@@ -169,7 +169,7 @@ subroutine SetServices (GC, RC)
 !   ------------------------
     call MAPL_GridCompSetEntryPoint (GC, ESMF_Method_Initialize,  Initialize, __RC__)
     call MAPL_GridCompSetEntryPoint (GC, ESMF_Method_Run, Run, __RC__)
-    if (data_driven /= .true.) then
+    if (data_driven .neqv. .true.) then
```
tclune, line 172: Am wondering if the following isn't clearer anyway given on of the operands is constant:
```
if (.not. data_driven) then
```
To my mind this is somewhat easier to grok.

tclune: And now that I see the other changes are duplicates of this,  it begs the introduction of a new logical variable:
```
logical :: not_data_driven
...
not_data_driven = .not. data_driven
...
if (not_data_driven) then
...
```

The question becomes, is there a better name than `not_data_driven`.  Ideally something more active/positive so that the "not" disappears entirely.   I'll ask anton.

tclune: I checked with Elliot, and the only antonym he could think of was `computational`.

Thinking ...

tclune: I suppose `model` gets the point across, but somehow it still does not sound right.

tclune: After a bit of googling, I think the term `numerical` is reasonably common:
```
logical :: numerical ! as opposed to data-driven
...
numerical = .not. data_driven
...
if (numerical) then
...
```

mathomp4:
Closing this as these must be in GOCART now. I mean, it built with GNU!

PR #16, opened 2020-10-27
Bug fix for reading bands from RC files

WilliamJamieson:
In running GOCART2G tests within NOAA's UFS system @rmontuoro discovered an error in the `PET.ESMF_LogFiles` related to reading the `BANDS:` option from the MAPL config.

This fixes this issue along with several related `rc` code errors found in GOCART2G's children.


adarmenov:
@WilliamJamieson - this PR is expected to be zero-diff, correct?

WilliamJamieson:
@adarmenov - this PR should be zero-diff. @gmao-esherman pointed out the default behavior for initializing the `MieTable` channels, I have now adjusted the logic to preserve the default behavior.

Note that there is a subtle bug that is being corrected. Namely, if a user had defined `BANDS:` in their config, current version of GOCART2G would not use this and instead overwrite their definitions with the defaults. This behavior has also been corrected.

gmao-esherman (approved):
Looks good to me. Thanks for catching this bug.

PR #20, opened 2021-02-23
Moving the recent changes to GOCART legacy into main

WilliamJamieson:
This copies the recent changes to GOCART legacy (the ones since this repository was setup) into the GOCART_GridComp. The majority of these changes appear to be the result of some minor Cmake changes to the GCM.

This is in preparation to move GOCART legacy in the main GCM from being part of the GEOSchem_GridComp to being cloned directly from this repository. I have already tested this along with these changes with a 1-day C12 run of the GCM and found the results to be zero-diff from the current version v10.17.3 of the GCM.


mathomp4:
Passed! :)

tclune (approved):

PR #22, opened 2021-02-23
Handmerge main into develop

mathomp4:
This is a handmerge to get `main` into `develop`. I might ping @WilliamJamieson to try this branch out. Maybe this was a bad idea...

PR #23, opened 2021-03-01
Compute diagnostic varaibles for multiple wavelengths. Stable.

gmao-esherman:
This is the latest stable branch for GOCART2G. The latest feature it includes is the ability to compute aerosol diagnostic variables at multiple wavelengths. Diagnostic wavelengths are defined in GOCART2G_GridComp.rc.


gmao-esherman:
I am not sure why this is failing on ci/circleci: build. How do I investigate this?

mathomp4:
@gmao-esherman I think this is because `develop` still has a bad circleCI configuration. I did fix it the PR from @WilliamJamieson that went into `main` but that isn't on `develop` until #22 can be pulled in...which of course doesn't build for us. 😦

gmao-esherman:
> @gmao-esherman I think this is because `develop` still has a bad circleCI configuration. I did fix it the PR from @WilliamJamieson that went into `main` but that isn't on `develop` until #22 can be pulled in...which of course doesn't build for us. 😦

Ahhh right... Thanks for the explanation. I guess we'll have to wait until those PR's go through.

adarmenov (commented):
1. What are these resources in BrC
> radius_lower: 0 0
> radius_upper: 0 0
> soil_density: 1800 1800

2. I am also not sure why AOP are calculated twice for NO3an1.   
3. Isn't the NH3 'bin' gas phase, and if so why is there a call to calculate AOP for it?

tclune:
@adarmenov If you don't know the answers to the questions you just posed, I'm going to guess that no one else in this ticket will either.   (@gmao-esherman might know some.)   Arlindo does not monitor github, so we may need to ask him in another medium.

gmao-esherman:
>     1. What are these resources in BrC
> 
> 
> > radius_lower: 0 0
> > radius_upper: 0 0
> > soil_density: 1800 1800
> 
>     1. I am also not sure why AOP are calculated twice for NO3an1.
> 
>     2. Isn't the NH3 'bin' gas phase, and if so why is there a call to calculate AOP for it?

1. The parameters you noted in CA2G_GridComp.br.rc (radius_lower, radius_upper, soil_density) should be removed. I will remove them. They currently do nothing, but were a holdover from and old version of the GA component.

2. I do not see where AOP's for NO3an1 are calculated twice. Are you referring to this portion of the code here? : https://github.com/GEOS-ESM/GOCART/blob/e4817708ecfa54f058940c197624b71f4d8630ad/ESMF/GOCART2G_GridComp/NI2G_GridComp/NI2G_GridCompMod.F90#L1064-L1090

If so, this broken into two calls because in legacy Nitrate only NO3an1 is used to calculate the PM2.5 variables. The other Nitrate AOP variables use NO3an1, NO3an2, and NO3an3. Please see legacy code here: https://github.com/GEOS-ESM/GEOSchem_GridComp/blob/695c8d1bd0453d3dd7bfb5a74a80b3326ee21300/GOCART_GridComp/NI_GridComp/NI_GridCompMod.F90#L1920-L1991

3. We only calculate certain diagnostic variables for NH3. NH3SMASS (surface mass), NH3CMASS (column integrated mass), NH3MASS (mass mixing ratio), and NH3CONC (mass concentration). This is what was done in legacy. https://github.com/GEOS-ESM/GEOSchem_GridComp/blob/3a061f570b78059415bc020599ec0bc6f4a88975/GOCART_GridComp/NI_GridComp/NI_GridCompMod.F90#L1874-L1895

adarmenov:
@gmao-esherman - yes, my second question was about the code in lines 1064-1065 and 1079-1082. If one of these calls (the first one?) is to calculate PM, is there a reason to provide the mie tables in the list of (optional?) arguments.

gmao-esherman:
> @gmao-esherman - yes, my second question was about the code in lines 1064-1065 and 1079-1082. If one of these calls (the first one?) is to calculate PM, is there a reason to provide the mie tables in the list of (optional?) arguments.

The only reason is that the mie table argument is not currently optional.

WilliamJamieson:
@gmao-esherman or @mathomp4  - Could you update this branch to the latest version of develop (there are some merge conflicts)?

ESMF/GOCART2G_GridComp/SU2G_GridComp/SU2G_GridCompMod.F90
```diff
@@ -1120,10 +1122,10 @@ subroutine Run2 (GC, import, export, clock, RC)
 
 !if(mapl_am_i_root()) print*,'SU2G Run2 WetRemoval sum(xh2o2) = ',sum(xh2o2)
 !if(mapl_am_i_root()) print*,'SU2G Run2 WetRemoval sum(self%h2o2_init) = ',sum(self%h2o2_init)
-
+!   Certain variables are multiplied by 1.0e-9 to convert from nanometers to meters
     call SU_Compute_Diags ( self%km, self%klid, self%radius(nSO4), self%sigma(nSO4), self%rhop(nSO4), &
                             MAPL_GRAV, MAPL_PI, nSO4, self%diag_MieTable(self%instance), &
-                            self%diag_MieTable(self%instance)%channels, &
+                            self%diag_MieTable(self%instance)%channels*1.0e-9, &
```
tclune, line 1128: I would not bother in this PR, but probably in the next changes, a named parameter like   NANOMETERS_PER_METER would be useful to make these factors of a billion more obvious to pure sw developers.

tclune (approved):

tclune:
Using admin superpowers to approve without GOCART codeowner approval.  Changes are all SW - not science.

ESMF/GOCART2G_GridComp/SU2G_GridComp/SU2G_GridCompMod.F90
gmao-esherman: This input is used in the call to Chem_MieQuery within SU_Compute_Diags. During today's meeting Arlindo ask that we change Chem_MieQuery to work in units of nanometers instead of meters. This would allow us to remove these types of unit conversions in the call.

adarmenov: We should stick to SI units in the code. Conversions like these are fine because they make the resource files more user friendly. What Tom suggested is to replace the hardcoded numerical value with a parameter/const variable to make the code more readable.

gmao-esherman: I do like the idea of sticking with SI units in the code. I thought Arlindo asked me to change Chem_MieQuery to work in nanometers in our meeting today? I must have misunderstood him.

adarmenov: My impression was that he requested a change of the 'monochromatic wavelengths' label, i.e. add 'nm' at the end of it, to make it similar to the other two 'wavelength' specific labels.

PR #24, opened 2021-03-01
NOAA UFS integration

rmontuoro:
This PR introduces code under development at NOAA aimed at integrating GOCART into the Unified Forecast System (UFS) weather model. It includes:

- A NUOPC-based interface implementing a UFS-compliant GOCART model component, in the new `ESMF/UFS` subdirectory
- Updates to the build system to build GOCART without `Baselibs` using standalone libraries, and to compile with 32-bit floating point variables

Note that the provided NUOPC/UFS cap enables MAPL to run under UFS, allowing for a potentially faster integration of components from the GEOS ecosystem into UFS in the future.


CMakeLists.txt
```diff
@@ -21,19 +21,52 @@ if (NOT CMAKE_BUILD_TYPE)
...
+option (BUILD_UFS "Build GOCART component for UFS" OFF)
+
 set (DOING_GEOS5 YES)
 
 # Should find a better place for this - used in Chem component
 set (ACG_FLAGS -v)
 set (FV_PRECISION R4R8)
 
+# Ensure we build as 32-bit
+message ("Force 32-bit build for GOCART")
```
mathomp4, line 34: @rmontuoro We don't really want people by fiat changing flags for GEOS. Is this something you can protect with `if (BUILD_UFS)`? Otherwise, it probably won't be accepted.

tclune: Ditto.

rmontuoro: Agreed.

ESMF/Aerosol_GridComp/AeroCinderella_GridComp/Cinderella/CMakeLists.txt
```diff
@@ -6,14 +6,13 @@ set (srcs
         ConvertUnits.F90
         NOAA_TracerMap.F90
         NOAA_GOCARTtracers.F90
-        MissingFields.F90
+#       MissingFields.F90
```
tclune, line 9: If MissingFields.F90 is no longer needed, should it be deleted?

ESMF/Aerosol_GridComp/Aerosol_GridComp.F90
```diff
@@ -43,6 +47,9 @@ subroutine SetServices(gc, rc)
         call ESMF_UserCompSetInternalState(gc, internal_name, wrap, status)
         VERIFY_(status)
 
+        ! add children
+        gocart = MAPL_AddChild(gc, name='GOCART2G', ss=GOCART2G_SetServices, __RC__)
+
```
tclune, line 52: OK - this one surprises me.  Didn't we add gocart as a child a long while back?  I'm clearly confused - just want to understand what I'm missing.

tclune (changes requested):
Just the one embedded question and request for modifying the 32-bit enforcer.

ESMF/Aerosol_GridComp/AeroCinderella_GridComp/Cinderella/CMakeLists.txt
rmontuoro: The Cinderella component is not used at all in this implementation. I generally agree with deleting unnecessary code, however I will defer to you for the decision. Note that this PR includes code under active development, so changes to the code structure are still possible.

ESMF/Aerosol_GridComp/Aerosol_GridComp.F90
rmontuoro: @tclune - Aerosol_GridComp is also not used/needed for the NOAA UFS integration

WilliamJamieson: @rmontuoro if it is not being used, then can it be removed? It is not being used in GEOS.

ESMF/Aerosol_GridComp/AeroCinderella_GridComp/Cinderella/CMakeLists.txt
WilliamJamieson: @rmontuoro Since the Cinderella component is no longer being used in this implementation, and it is not being used by GEOS then it can be eliminated also. @tclune do you agree?

ESMF/Apps/CMakeLists.txt
```diff
@@ -1,13 +1,13 @@
-ecbuild_add_executable(
-        TARGET SyntheticApp.x
-        SOURCES SyntheticApp.F90 Synthetic_Driver.F90 Provider_GridComp.F90 Reciever_GridComp.F90 UFS_TestingCap.F90
-        LIBS gftl-shared MAPL MAPL.NUOPC
-        INCLUDES ${esma_include}/MAPL
-)
+add_executable(SyntheticApp.x
```
WilliamJamieson, line 1: @rmontuoro Are any of the `Apps` used for your UFS work? If not then they can be deleted as they were just used for development testing

ESMF/Aerosol_GridComp/Aerosol_GridComp.F90
gmao-esherman: Aerosol_GridComp is likely used in GEOS for the carbon and CFC components in legacy GOCART.

ESMF/Apps/CMakeLists.txt
rmontuoro: @WilliamJamieson - `Apps` are not being used for UFS development work. I'm fine with deleting the subdirectory if you agree.

ESMF/Aerosol_GridComp/Aerosol_GridComp.F90
WilliamJamieson: @gmao-esherman, This is incorrect, this component was introduced as the original "root" component for the effort to allow GOCART2G to be used in the UFS. The development has now moved away from that.

ESMF/GOCART2G_GridComp/SS2G_GridComp/SS2G_GridCompMod.F90
```diff
@@ -842,9 +842,11 @@ subroutine Run2 (GC, import, export, clock, RC)
                              SSFLUXU, SSFLUXV, SSCONC, SSEXTCOEF, SSSCACOEF,    &
                              SSEXTTFM, SSSCATFM ,SSANGSTR, SSAERIDX, NO3nFlag=.false.,__RC__)
  
-do n=1,5
-   if(mapl_am_i_root()) print*,'n = ', n,' : Run2 E SS2G sum(ss00n) = ',sum(SS(:,:,:,n))
-end do
+    if (mapl_am_i_root()) then
+      do n = 1, size(SS, 4)
+        print*,'n = ', n,' : Run2 E SS2G sum(ss00n) = ',sum(SS(:,:,:,n))
```
WilliamJamieson, line 847: @gmao-esherman and @rmontuoro, this appears to be printing something every time-step. Should we still be doing this?

ESMF/Apps/CMakeLists.txt
WilliamJamieson: @rmontuoro I am fine with doing this as well.

ESMF/Aerosol_GridComp/Aerosol_GridComp.F90
gmao-esherman: My mistake. I was referring to Aero_GridComp.

tclune:
Yes - in fact was trying to ping you in Teams when I saw your GitHub response.   If NOAA is confident they won't use it, GEOS certainly does not need it.

ESMF/GOCART2G_GridComp/SS2G_GridComp/SS2G_GridCompMod.F90
rmontuoro: @WilliamJamieson - Print statements like the one you pointed out, if needed, could be turned on/off based on global debug level settings for GOCART. However, it would be more helpful if debug output could be sent to ESMF Log files, as  it will provide information for all PETs, not just root. In any case, the code is not necessary in UFS.

gmao-esherman: There are many prints like this in all of GOCART2G's aerosol components. They are not needed anymore.

tclune:
If it is needed in some debugging scenarios, it could/should be converted to pflogger calls.

WilliamJamieson:
> If it is needed in some debugging scenarios, it could/should be converted to pflogger calls.

You are referring to all these "debugging" prints throughout GOCART2G?

WilliamJamieson:
@rmontuoro we just merged some changes onto `develop` could you merge these changes onto this branch and check that it still works?

CMakeLists.txt
rmontuoro: Done. See commit d06f07d

ESMF/Aerosol_GridComp/AeroCinderella_GridComp/Cinderella/CMakeLists.txt
rmontuoro: Commit 195f02c removed the Cinderella component.

ESMF/Apps/CMakeLists.txt
rmontuoro: Commit f8c2208 removed test `Apps`

ESMF/Aerosol_GridComp/Aerosol_GridComp.F90
rmontuoro: I take that back. Aerosol_GridComp *is* needed to initialize  and run MAPL Cap in the NUOPC/UFS cap:

```
  use Aerosol_GridComp_mod, only : &
    aerosol_routine_SS          => SetServices
  [...]
  is % maplCap = MAPL_Cap("MAPL Driver", aerosol_routine_SS, cap_options=maplCapOptions, _RC)
```

@tclune - Are additional changes required to correctly set up the MAPL component hierarchy?

tclune: This aspect is slightly unstable in the midst of my other recent refactorings and thus a bit hard to assess from just a quick glance at the code.    Generally if anything is wrong wrt to the MAPL hierarchy, things will break immediately.   So "working for you" is a sufficient indicator, and all that I can offer at the moment.   Of course if you do see a problem, we'll jump in immediately to diagnose.

CMakeLists.txt
```diff
@@ -21,19 +21,54 @@ if (NOT CMAKE_BUILD_TYPE)
...
 if (NOT COMMAND esma)
     list (APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake@")
     include (esma)
 endif()
 
     ecbuild_declare_project()
 
+# Find dependencies
+if (NOT NetCDF_Fortran_FOUND)
+  find_package (NetCDF REQUIRED Fortran)
```
WilliamJamieson, line 52: These `'find_package` calls are causing cmake errors for me when I try to build GEOS using this branch. They should be wrapped in a `if (BUILD_UFS)` or `if (NOT Baselibs_FOUND)`? @mathomp4 suggested that it maybe best to have both.

WilliamJamieson (changes requested):

CMakeLists.txt
mathomp4: Yeah. My thought was the `find_package` commands could be in an `if (NOT Baselibs_FOUND)` block as is done in MAPL (https://github.com/GEOS-ESM/MAPL/blob/main/CMakeLists.txt).

Meanwhile the flag options could be in a more UFS specific if-block

rmontuoro: I agree with @mathomp4 - we can add a `if (NOT Baselibs_FOUND)` block wrapping all Baselibs-specific `find_package` commands.

The flag option has been wrapped with `if (BUILD_UFS)` in commit d06f07d0.

rmontuoro: @mathomp4, @WilliamJamieson - could you please provide directions on how to build GOCART with Baselibs, as done in GEOS? I'd like to test this change before submitting it.

mathomp4: @rmontuoro I think we could help. Baselibs is (usually nowadays) pretty simple to build, though every new machine seems to throw its own special wrench. But a set up of GNU and Open MPI should get you most of the way there (maybe MKL or BLAS/LAPACK is needed for GEOS? MKL is free, so I often forget about it).

rmontuoro: Commit a9c1314 wraps all `find_package` calls with a `if (NOT Baselibs_FOUND)` block, as agreed.

After these changes, GOCART builds successfully on Orion both within UFS-GOCART and following @WilliamJamieson's [instructions](https://github.com/GEOS-ESM/GOCART/wiki/Building-GOCART-on-Orion ), which use Baselibs. Note, however, that in the latter case the included legacy GOCART (`GOCART_GridComp`) did not build since it requires additional chemistry modules. Also, MAPL should be added as an explicit dependency in the `CMakeLists.txt` file for each of the legacy GOCART components. For example:

```
diff --git a/ESMF/GOCART_GridComp/BC_GridComp/CMakeLists.txt b/ESMF/GOCART_GridComp/BC_GridComp/CMakeLists.txt
index a26a18c..bfb9224 100644
--- a/ESMF/GOCART_GridComp/BC_GridComp/CMakeLists.txt
+++ b/ESMF/GOCART_GridComp/BC_GridComp/CMakeLists.txt
@@ -2,7 +2,7 @@ esma_set_this ()
 
 esma_add_library (${this}
   SRCS ${this}Mod.F90
-  DEPENDENCIES Chem_Shared Chem_Base GMAO_mpeu MAPL_cfio_r4)
+  DEPENDENCIES Chem_Shared Chem_Base GMAO_mpeu MAPL MAPL_cfio_r4)
 target_include_directories (${this} PUBLIC ${INC_ESMF} ${INC_NETCDF})
```

tclune (dismissed):
We need to freeze this PR.    It is very big and every time a change is made we need to review it again.

CMakeLists.txt
WilliamJamieson: @rmontuoro this is an excellent point. I will look into making these changes.

CMakeLists.txt
```diff
@@ -21,19 +21,56 @@ if (NOT CMAKE_BUILD_TYPE)
...
+
+  find_package (GFTL REQUIRED)
+  find_package (GFTL_SHARED REQUIRED)
+  find_package (PFLOGGER REQUIRED)
+  find_package (YAFYAML REQUIRED)
+endif ()
+
+if (IS_DIRECTORY "${PROJECT_SOURCE_DIR}/ESMF/Shared/MAPL@")
+  message (STATUS "Using local MAPL source tree: ${PROJECT_SOURCE_DIR}/ESMF/Shared/MAPL@")
+else ()
```
WilliamJamieson, line 69: Another build error, this time for building with the full `GEOSgcm`. Could you change this line to:
```
elseif (NOT Baselibs_Found)
```
This will prevent the find package from interfering with the `GEOSgcm` build.

WilliamJamieson (changes requested):

CMakeLists.txt
WilliamJamieson: Note that with this change, I can confirm that this branch will build and zero-diff run GOCART legacy with the `GEOSgcm`.

rmontuoro: @WilliamJamieson - I'm a bit confused. MAPL is not bundled with Baselibs. Does `GEOSgcm` build places MAPL in a different directory? If the directory is the same (`GOCART/ESMF/Shared/MAPL@`), the current `CMakeLists.txt` should work (it worked on Orion).

WilliamJamieson: MAPL goes in a completely different directory when in the `GEOSgcm` thats the issue.

WilliamJamieson: @rmontuoro - This is fine with me.

WilliamJamieson: This appears to have been addressed by commit 5f69e7c.

WilliamJamieson (approved):
The latest commit appears to have resolved all my concerns.

gmao-esherman (approved):
I have verified this does not break GOCART2G, and is zero-diff.

PR #31, opened 2021-03-04
Changing `stop` to non-zero rc in process library 

### LINKED ISSUE #30: Process library APIs must return error status
GOCART2G process library procedures must return an error status to allow the calling application to gracefully handle errors.

They also should not include `stop` statements, which may cause the application to hang.

WilliamJamieson:
This should fix #30, by changing all the procedures which have `stop` statements to a non-zero return code.


Process_Library/GOCART2G_Process.F90
```diff
@@ -1,10 +1,13 @@
 #define __SUCCESS__ 0
+#define __FAIL__ 1
 #define __VERIFY__(x) if(x/=0) then; if(present(rc)) rc=x; return; endif
 #define __RC__ rc=status); __VERIFY__(status
 #define __STAT__ stat=status); __VERIFY__(status
 #define __IOSTAT__ iostat=status); __VERIFY__(status
 #define __RETURN__(x) if (present(rc)) rc=x; return
 #define __ASSERT__(expr) if(.not. (expr)) then; if (present(rc)) rc=-1; return; endif
+#define __VERIFY_NO_OPT__(x) if(x/=0) then; rc=x; return; endif
+#define __RC_NO_OPT__ rc=status); __VERIFY_NO_OPT__(status
```
tclune, line 10: If I find other changes, I recommend moving the new variants to be right next to their originals.   Will make maintenance easier if something needs to change.  But not worth it if this is the only change.

Process_Library/GOCART2G_Process.F90
```diff
@@ -8360,7 +8420,10 @@ subroutine PrintError  &
       Write (6,*)
 
       if (err_do_stop) then
-        stop "Code stopped by PrintError."
+        ! TODO: fix stop statement
+        ! stop "Code stopped by PrintError."
```
tclune, line 8424: I think there are a lot of comments that should be deleted.  In particular these.

tclune (changes requested):
The good thing is that with the macros, the compiler should catch the most obvious mistakes involving optional/nonoptional.

Main concern is lots of comments that look like they should be deleted.

WilliamJamieson:
> The good thing is that with the macros, the compiler should catch the most obvious mistakes involving optional/nonoptional.
> 
> Main concern is lots of comments that look like they should be deleted.

I left the comments in for now in case we wanted to discuss them further. I'll go ahead and delete them.

Process_Library/GOCART2G_Process.F90
WilliamJamieson: Fixed by comment 28920daf5abe068caf4a48985da15e5ca1a61e8b.

Process_Library/GOCART2G_Process.F90
WilliamJamieson: Fixed by commit 28920daf5abe068caf4a48985da15e5ca1a61e8b.

Process_Library/GOCART2G_Process.F90
```diff
@@ -2714,14 +2719,15 @@ subroutine hoppelCorrection (radius, rhop, rh, dz, ustar, rhFlag, &
...
-   rc = 0
+   rc = __SUCCESS__
    fhoppel = 1.0
    allocate(vsettle, mold=rh)
 
    do j = 1, ubound(rh,2)
       do i = 1, ubound(rh,1)
          call wetRadius (radius, rhop, rh(i,j), rhFlag, &
                          radius_wet, rhop_wet, rc)
+         ! TODO: wetRadius can never return a non-zero rc so this statement is useless
```
WilliamJamieson, line 2730: I left this comment in because what was going on here is very strange.

tclune (dismissed):

Process_Library/GOCART2G_Process.F90
WilliamJamieson: Just fixed this issue

WilliamJamieson:
@tclune - The return code checking and handling within the `Chem_MieTableMod2G.F90` file seems to be inconsistent. Should I take a closer look?

tclune:
Can you elaborate on the MieTable issue?    It looks like all the procedures have an `rc` argument.  Most are optional, one is required.    I did not look at the `NF_VERIFY_` macro, but so long as it returns when rc is nonzero, I'm not seeing a problem.

tclune (approved):

PR #35, opened 2021-03-08
Fixes #26. Update CODEOWNERS

### LINKED ISSUE #26: Repository needs updated CODEOWNERS
This repo at present does not have a `CODEOWNERS` file. I can add one, I just need to know who to put in it. I mean, I guess @gmao-esherman should be there in some way. 

I suppose it should reflect the [GEOSchem_GridComp CODEOWNERS](https://github.com/GEOS-ESM/GEOSchem_GridComp/blob/develop/.github/CODEOWNERS) file. But should I make a new "GOCART2G Team"?

ETA: Actually, there was a `CODEOWNERS` just not in `main` at the time. There is one in `develop` at least.

mathomp4:
Update the `CODEOWNERS` to reflect latest repository.

NOTE: This is a draft until I get feedback from people more in the know. For example, we might want to make a GOCART2G Team until that is "live"?

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -6,12 +6,12 @@
 # Everything in GOCART should be owned by the Chem gatekeepers
 * @GEOS-ESM/chemistry-gatekeepers
 
-# NOAA Extras are Tom, William for now
-/CCPP/                  @tclune @weiyuan_jiang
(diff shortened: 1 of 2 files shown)


tclune (changes requested):
We should remove William while we are thinking about it.

tclune (approved):

PR #36, opened 2021-03-08
GOCART errors

WilliamJamieson:
@gmao-esherman reported two errors resulting from the `rc` handling PR #31, due to it catching non-zero return codes. Namely, the tracebacks:
```
pe=00010 FAIL at line=00832 NI2G_GridCompMod.F90 <status=-858993460>
pe=00025 FAIL at line=01092 SU2G_GridCompMod.F90 <status=100>
```
The first was cased by an uninitialized value for the `rc` in the process library. While the second is due to the inclusion of a gas phase in the SU bins, so there shouldn't be any settling for that bin anyways.


tclune (dismissed):

gmao-esherman (approved):
Changes look good to me. I have also tested them in my sanbox.

PR #37, opened 2021-03-10
Fixing more GOCART rc related errors

WilliamJamieson:
@gmao-esherman Noted another rc error, which was the result of an uninitialized `rc` in the process library.

Thanks to @atrayano who spotted where to initialize the value.

I have tested this with GOCART2G enabled at C12 on discover, and it fixes the error.


WilliamJamieson:
Merged to main not develop. grr

PR #40, opened 2021-03-22
Compute and export total PM2.5 and PM10 for UFS

rmontuoro:
This PR introduces changes to the UFS/NUOPC cap to compute and export total diagnostic PM2.5 and PM10 for the UFS weather model. These quantities are computed according to formulas provided by @amdasilva.

The changes have been tested within UFS-GOCART on NOAA's Hera platform.

### DIFF
--- a/ESMF/UFS/Aerosol_Comp_Mod.F90
+++ b/ESMF/UFS/Aerosol_Comp_Mod.F90
@@ -41,28 +41,33 @@ module Aerosol_Comp_Mod
   real(ESMF_KIND_R8), parameter :: ppm2one = 1.e-06_ESMF_KIND_R8
 
   ! tracer map
+  ! - prognostic section (advected)
   integer, parameter :: p_o3     = 7
   integer, parameter :: p_so2    = p_o3 + 1
   integer, parameter :: p_sulf   = p_o3 + 2
   integer, parameter :: p_dms    = p_o3 + 3
   integer, parameter :: p_msa    = p_o3 + 4
-  integer, parameter :: p_bc_1   = p_o3 + 6
-  integer, parameter :: p_bc_2   = p_o3 + 7
-  integer, parameter :: p_oc_1   = p_o3 + 8
-  integer, parameter :: p_oc_2   = p_o3 + 9
-  integer, parameter :: p_dust_1 = p_o3 + 10
-  integer, parameter :: p_dust_2 = p_o3 + 11
-  integer, parameter :: p_dust_3 = p_o3 + 12
-  integer, parameter :: p_dust_4 = p_o3 + 13
-  integer, parameter :: p_seas_1 = p_o3 + 15
-  integer, parameter :: p_seas_2 = p_o3 + 16
-  integer, parameter :: p_seas_3 = p_o3 + 17
-  integer, parameter :: p_seas_4 = p_o3 + 18
-  integer, parameter :: p_nh3    = p_o3 + 20
-  integer, parameter :: p_nh4a   = p_o3 + 21
-  integer, parameter :: p_no3an1 = p_o3 + 22
-  integer, parameter :: p_no3an2 = p_o3 + 23
-  integer, parameter :: p_no3an3 = p_o3 + 24
+  integer, parameter :: p_bc_1   = p_o3 + 5
+  integer, parameter :: p_bc_2   = p_o3 + 6
+  integer, parameter :: p_oc_1   = p_o3 + 7
+  integer, parameter :: p_oc_2   = p_o3 + 8
+  integer, parameter :: p_dust_1 = p_o3 + 9
+  integer, parameter :: p_dust_2 = p_o3 + 10
+  integer, parameter :: p_dust_3 = p_o3 + 11
+  integer, parameter :: p_dust_4 = p_o3 + 12
+  integer, parameter :: p_seas_1 = p_o3 + 14
+  integer, parameter :: p_seas_2 = p_o3 + 15
+  integer, parameter :: p_seas_3 = p_o3 + 16
+  integer, parameter :: p_seas_4 = p_o3 + 17
+  integer, parameter :: p_nh3    = p_o3 + 19
+  integer, parameter :: p_nh4a   = p_o3 + 20
+  integer, parameter :: p_no3an1 = p_o3 + 21
+  integer, parameter :: p_no3an2 = p_o3 + 22
+  integer, parameter :: p_no3an3 = p_o3 + 23
+  ! - diagnostic section (not advected)
+  integer, parameter :: p_pm25   = p_o3 + 24
+  integer, parameter :: p_pm10   = p_o3 + 25
+
   integer, parameter :: p_seas_d = p_seas_1 - p_so2 - 3
 
   interface AerosolGetPtr
--- a/ESMF/UFS/Aerosol_Comp_Mod.F90
+++ b/ESMF/UFS/Aerosol_Comp_Mod.F90
@@ -51,7 +51,13 @@ module Aerosol_Comp_Mod
   integer, parameter :: p_oc_1   = p_o3 + 8
   integer, parameter :: p_oc_2   = p_o3 + 9
   integer, parameter :: p_dust_1 = p_o3 + 10
+  integer, parameter :: p_dust_2 = p_o3 + 11
+  integer, parameter :: p_dust_3 = p_o3 + 12
+  integer, parameter :: p_dust_4 = p_o3 + 13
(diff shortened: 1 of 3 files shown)


mathomp4:
Whoops! Looks like I need to make some surgery on the CI. Give me a moment or two...

mathomp4:
@rmontuoro I'm throwing a blocker label on this until I get the CI fixed. That'll need a pull from `main` due to #41. Note: this will also add the Changelog enforcer (like you've seen in MAPL). I've been lax in adding that.

ESMF/UFS/Aerosol_Comp_Mod.F90
```diff
@@ -40,6 +40,35 @@ module Aerosol_Comp_Mod
...
+  integer, parameter :: p_seas_3 = p_o3 + 16
+  integer, parameter :: p_seas_4 = p_o3 + 17
+  integer, parameter :: p_nh3    = p_o3 + 19
+  integer, parameter :: p_nh4a   = p_o3 + 20
+  integer, parameter :: p_no3an1 = p_o3 + 21
+  integer, parameter :: p_no3an2 = p_o3 + 22
+  integer, parameter :: p_no3an3 = p_o3 + 23
+  ! - diagnostic section (not advected)
+  integer, parameter :: p_pm25   = p_o3 + 24
+  integer, parameter :: p_pm10   = p_o3 + 25
```
tclune, line 69: perhaps consider using a Fortran enumerator instead?

tclune: Ah - I now see this was just a move.   But probably still worth considering.

tclune (dismissed):

gmao-esherman (dismissed):

ESMF/UFS/Aerosol_Comp_Mod.F90
```diff
@@ -40,6 +40,35 @@ module Aerosol_Comp_Mod
...
+
+  integer, parameter :: p_seas_d = p_seas_1 - p_so2 - 3
```
adarmenov, line 71: I do not follow... what is the connection between seasalt and SO2?

ESMF/UFS/Aerosol_Comp_Mod.F90
```diff
@@ -709,6 +720,101 @@ subroutine AerosolTracerUpdate(state, cap, rc)
...
+      kk = nk - k + 1
+      do j = 1, nj
+        do i = 1, ni
+          ! -- compute partial PM2.5
+          pm25 = q(i,j,k,p_bc_1) + q(i,j,k,p_bc_2)     &
+               + q(i,j,k,p_oc_1) + q(i,j,k,p_oc_2)     &
+               + q(i,j,k,p_dust_1)                     &
+               + q(i,j,k,p_seas_1) + q(i,j,k,p_seas_2) &
+               + w_so4 * q(i,j,k,p_sulf)               &
+               + w_no3 * q(i,j,k,p_no3an1)
```
adarmenov, line 795: 1. @gmao-esherman Would this work if one or more aerosol types are disabled, dust for example?
2. @gmao-esherman I would expect that the PM diagnostics is calculated using only the active aerosols, is this the case?

ESMF/UFS/Aerosol_Comp_Mod.F90
```diff
@@ -709,6 +720,101 @@ subroutine AerosolTracerUpdate(state, cap, rc)
...
+
+    ! -- local parameters
+    real(ESMF_KIND_R8), parameter :: w_du2  = log(1.250_ESMF_KIND_R8) / log(1.8_ESMF_KIND_R8)
+    real(ESMF_KIND_R8), parameter :: w_du4  = log(1.667_ESMF_KIND_R8) / log(2.0_ESMF_KIND_R8)
+    real(ESMF_KIND_R8), parameter :: w_ss3  = log(2.50_ESMF_KIND_R8) / log(3._ESMF_KIND_R8)
+    real(ESMF_KIND_R8), parameter :: w_so4  = 132.14_ESMF_KIND_R8 / 96.06_ESMF_KIND_R8
+    real(ESMF_KIND_R8), parameter :: w_no3  = 80.043_ESMF_KIND_R8 / 62.0_ESMF_KIND_R8
+    real(ESMF_KIND_R8), parameter :: w25_no3an2 = 0.138_ESMF_KIND_R8 * w_no3
+    real(ESMF_KIND_R8), parameter :: w10_no3an2 = 0.808_ESMF_KIND_R8 * w_no3
+    real(ESMF_KIND_R8), parameter :: w10_no3an3 = 0.164_ESMF_KIND_R8 * w_no3
```
adarmenov, line 746: Is r8 really necessary?

ESMF/UFS/Aerosol_Comp_Mod.F90
gmao-esherman: This is all under ESMF/UFS (NOAA stuff). Nothing under ESMF/UFS gets exercised by legacy GOCART or GOCART2G when we build GEOS.

adarmenov: @gmao-esherman  Why are we asked to review it then, just a formality?

gmao-esherman: Good question. I am not 100% sure why. I think any changes made to the @GOCART repository have to get reviewed by someone. Maybe @tclune has some more insight.

tclune:
@adarmenov    I'm trying to fix it so you won't have to review.    Currently the CODEOWNERS file has you and Elliot reviewing changes anywhere under GOCART.   This one file should be managed separately.        Hopefully soon, but you just know it will happen after the last change needed by NOAA.

mathomp4:
> @adarmenov I'm trying to fix it so you won't have to review. Currently the CODEOWNERS file has you and Elliot reviewing changes anywhere under GOCART. This one file should be managed separately. Hopefully soon, but you just know it will happen after the last change needed by NOAA.

Huh. How did @adarmenov get tagged on this. The CODEOWNER says:
```
/ESMF/UFS/              @tclune @weiyuan-jiang
```
Maybe I got this wrong...

ESMF/UFS/Aerosol_Comp_Mod.F90
rmontuoro: @tclune - agreed. This should not be hardcoded and will hopefully be overhauled soon, perhaps using GFTL maps.
Are enumerators fully supported across Fortran compilers?

tclune:
I don't use enumerators much, but the C-interoperability stuff is "easy" and I've not heard of any bug reports in that direction.    F202x will hopefully have a much more useful type of enumerator, as these are rather limited.

tclune:
Certainly they are supported by all compilers that support current MAPL.

ESMF/UFS/Aerosol_Comp_Mod.F90
rmontuoro: @adarmenov - `p_seas_d` is a computed offset to populate UFS-defined sea salt diagnostics. The only connection between sea salt and SO2 is their relative order in the tracer array.

ESMF/UFS/Aerosol_Comp_Mod.F90
rmontuoro: @adarmenov - no. However, PM2.5 and PM10 are computed using tracer mixing ratios stored in an `ESMF_TYPEKIND_R8` coupling field, so `ESMF_KIND_R8` is being used for consistency.

ESMF/UFS/Aerosol_Comp_Mod.F90
rmontuoro: @adarmenov, @gmao-esherman - Good point. As long as all required aerosol tracers are included in the atmospheric model, the PM computation will not be affected by one or more aerosol types being disabled. More flexibility might be added in the future.

ESMF/UFS/Aerosol_Comp_Mod.F90
adarmenov: I can see that... This is very fragile in my opinion. Consider also the magical number 3, which I guess is the equivalent of (p_msa - p_so2).

rmontuoro: @adarmenov - all valid points. This code is slated to be refactored and diagnostics may be removed as implemented. The number 3 provides the offset for the exported diagnostic array, and it is not related to (p_msa-p_so2).

gmao-esherman (approved):

tclune (approved):

PR #45, opened 2021-03-23
Update CMakeLists.txt

mathomp4:
For some reason, the GOCART CMakeLists.txt was setting `FV_PRECISION`. I can see no need for this! 

Also, clean up indent.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,7 +28,6 @@ set (DOING_GEOS5 YES)
 
 # Should find a better place for this - used in Chem component
 set (ACG_FLAGS -v)
-set (FV_PRECISION R4R8)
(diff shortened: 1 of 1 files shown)


mathomp4:
Adding @rmontuoro to this in case he knows of a reason to set `FV_PRECISION` in GOCART. Is UFS perhaps depending on it?

tclune (approved):

rmontuoro:
@mathomp4 - UFS does not use `FV_PRECISION`. Thank you for tidying up GOCART `CMakeLists.txt`

mathomp4:
Changing to 0-diff-trivial since I don't think we need to fill up the Changelog with this :)

PR #52, opened 2021-04-07
Feature/esherman/gocart2g develop v2

gmao-esherman:
This PR:

1. Removes diagnostic print statements.
2. Fixes bug with the creation of the Diagnostic Mie tables.
3. Adds callbacks needed for GAAS2G.


tclune (approved):
I recommend following this up with a PR to update the changelog to indicate the current status.  Does not need to be detailed - just a high level summary.

ESMF/GOCART2G_GridComp/CA2G_GridComp/CA2G_GridCompMod.F90
```diff
@@ -1295,6 +1278,172 @@ end subroutine mie_
...
+    call ESMF_AttributeGet(state, name='mieTable_pointer', valueList=opaque_self, __RC__)
+
+    address = transfer(opaque_self, address)
+    call c_f_pointer(address, self)
+
+    do n = 1, nbins
+      do i = 1, i2
+        do j = 1, j2
+          do k = 1, km
+            call Chem_MieQuery(self%diag_MieTable(instance), n, mieTable_index, q_4d(i,j,k,n), rh(i,j,k), tau(i,j,k), __RC__)
```
mathomp4, line 1393: As an aside, I remember back in the old day I created `Chem_MieQueryAllBand4D`(and `3D`) in `Chem_MieMod.F90` because it was quite expensive to make a MieQuery inside of 3 or 4 loops. So I moved the do loops inside the query so I could call it *once* or at most `nbands` times. 

You might want to look at something similar here, though maybe the compilers then were just bad at this? @tclune any idea if this still might be the case in Fortran compilers?

mathomp4: ETA: actually it's called once. The difference between 3D and 4D is that some places called it with (i,k,nbands) and other (i,j,k,nbands)

gmao-esherman: We made this an 'impure elemental' subroutine to get rid of the nested loops, but I was having problems getting it to work which is why the nested loops are still there. This is something I can revisit in the future.

tclune (approved):

tclune:
Indeed - we had looked at making this interface elemental precisely to avoid the looping.  Not sure of the full history though.  @gmao-esherman

gmao-esherman:
> Indeed - we had looked at making this interface elemental precisely to avoid the looping. Not sure of the full history though. @gmao-esherman

Correct. The interface is 'impure elemental', but I still couldn't get it work. For the sake of time and effort, I used the nested loops. Interestingly it does work as an elemental interface with the aerosol_optics subroutines. This is something to revisit.

gmao-esherman:
It looks like the build failed in GAAS:
/root/project/GEOSgcm/src/Components/@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/@GEOSchem_GridComp/GAAS_GridComp/GAAS_GridCompMod.F90:339:14:

  339 |    self%y_f = MAPL_SimpleBundleCreate (self%aodGrid, rc=status, name='y_f')

I don't understand why it's failing here. The changes to GAAS are in the @GEOSchem_GridComp repo, and I also merged the MAPL_SimpleBundleCreate PR into @MAPL develop at least a week or two ago.

mathomp4:
> It looks like the build failed in GAAS:
> /root/project/GEOSgcm/src/Components/@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/@GEOSchem_GridComp/GAAS_GridComp/GAAS_GridCompMod.F90:339:14:
> 
> 339 | self%y_f = MAPL_SimpleBundleCreate (self%aodGrid, rc=status, name='y_f')
> 
> I don't understand why it's failing here. The changes to GAAS are in the @GEOSchem_GridComp repo, and I also merged the MAPL_SimpleBundleCreate PR into @mapl develop at least a week or two ago.

@gmao-esherman Ah. The issue is that while GOCART `develop` points to MAPL `develop`, GEOSgcm does not. It points to fixed tags and in this case that is MAPL 2.6.4.

I suppose the question is: should the CI in GOCART when it does a GEOSgcm build test, do an additional `mepo develop MAPL`? Since by definition (at the moment) GOCART2G requires it? @tclune Thoughts?

Process_Library/GOCART2G_Process.F90
```diff
@@ -6654,7 +6654,7 @@ subroutine get_HenrysLawCts(name,c1,c2,c3,c4,rc)
    end subroutine get_HenrysLawCts
 
 !==================================================================================
-!BOP
+!BOh
```
mathomp4, line 6657: Note: This will break protex...if it currently worked. Probably good to return it to `!BOP`

ESMF/GOCART_GridComp/NI_GridComp/NI_GridCompMod.F90
```diff
@@ -1343,18 +1343,26 @@ subroutine NI_GridCompRun2_ ( gcNI, w_c, impChem, expChem, &
    if(associated(NI_pnh4aq%data2d)) NI_pnh4aq%data2d(:,:) = 0.
    if(associated(NI_pnh3aq%data2d)) NI_pnh3aq%data2d(:,:) = 0.
 
+if(mapl_am_i_root()) print*,'NI before thermo sum(NH3) = ',sum(w_c%qa(nNH3)%data3d)
```
mathomp4, line 1346: @gmao-esherman Does GEOS GOCART legacy run NI? If so, this will print a lot on each GOCART time step, will it not? Perhaps this should be pflogger based?

mathomp4:
Wait. Huh. Why did that GAAS fail?

mathomp4:
> Wait. Huh. Why did that GAAS fail?

Ahhh. Now I see. this is mepo being clever. It found the `feature/esherman/gocart2g_develop_v2` branch of GEOSchem_GridComp.

My suggestion, I guess would be to create a branch off of MAPL `develop` with the same name. Then when mepo runs `mepo checkout-if-exists feature/esherman/gocart2g_develop_v2` it finds a MAPL branch of the same name.

tclune:
I don't think so.   GCM should not be using `develop` from MAPL.  Thus for 2G to go into GCM, we need to do a release of MAPL and put that into the GCM prior to merging 2G into main.

mathomp4:
oooh. Huh. Okay. This is interesting. Yeah, the GCM CI is dumber than you might be thinking. The issue is that the CI will only ever do `git clone GEOSgcm.git` and get the main. It only looks for matching branches *AFTER* it's cloned main. Thus, even though you have a `feature/esherman/gocart2g_develop_v2` branch on GEOSgcm, `mepo clone` already grabbed MAPL 2.6.4.

After that, it just runs `mepo checkout-if-exists BRANCH` where BRANCH is the branch here in GOCART.  Your branch on GEOSgcm has:
```diff
diff --git a/components.yaml b/components.yaml
index 65e626a..269bdba 100644
--- a/components.yaml
+++ b/components.yaml
@@ -5,7 +5,7 @@ GEOSgcm:
 env:
   local: ./@env
   remote: ../ESMA_env.git
-  tag: v3.2.0
+  tag: v3.1.3
   develop: main

 cmake:
@@ -29,14 +29,14 @@ NCEP_Shared:
 GMAO_Shared:
   local: ./src/Shared/@GMAO_Shared
   remote: ../GMAO_Shared.git
-  tag: v1.3.10
+  tag: v1.3.9
   sparse: ./config/GMAO_Shared.sparse
   develop: main

 MAPL:
   local: ./src/Shared/@MAPL
   remote: ../MAPL.git
-  tag: v2.6.4
+  branch: develop
   develop: develop

 FMS:
@@ -48,33 +48,33 @@ FMS:
 GEOSgcm_GridComp:
   local: ./src/Components/@GEOSgcm_GridComp
   remote: ../GEOSgcm_GridComp.git
-  tag: v1.11.6
+  branch: feature/esherman/gocart2g_develop_WilliamUpdate
   sparse: ./config/GEOSgcm_GridComp.sparse
   develop: develop

 FVdycoreCubed_GridComp:
   local: ./src/Components/@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSsuperdyn_GridComp/@FVdycoreCubed_GridComp
   remote: ../FVdycoreCubed_GridComp.git
-  tag: v1.2.11
+  tag: v1.2.10
   develop: develop

 fvdycore:
   local: ./src/Components/@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSsuperdyn_GridComp/@FVdycoreCubed_GridComp/@fvdycore
   remote: ../GFDL_atmos_cubed_sphere.git
-  tag: geos/v1.1.5
+  tag: geos/v1.1.4
   develop: geos/develop

 GEOSchem_GridComp:
   local: ./src/Components/@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/@GEOSchem_GridComp
   remote: ../GEOSchem_GridComp.git
-  tag: v1.4.4
+  branch: feature/esherman/gocart2g_develop_v2
   develop: develop

 GOCART:
   local: ./src/Components/@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/@GEOSchem_GridComp/@GOCART
   remote: ../GOCART.git
-  sparse: ./config/GOCART.sparse
-  tag: v1.0.1
+#  sparse: ./config/GOCART.sparse
+  branch: feature/esherman/gocart2g_develop_v2
   develop: develop

 mom:
@@ -86,14 +86,14 @@ mom:
 mom6:
   local: ./src/Components/@GEOSgcm_GridComp/GEOSogcm_GridComp/GEOSocean_GridComp/GuestOcean_GridComp/MOM6_GEOSPlug/@mom6
   remote: ../MOM6.git
-  tag: geos/v2.0.1
+  tag: geos/v2.0.0
   develop: dev/gfdl
   recurse_submodules: True

 GEOSgcm_App:
   local: ./src/Applications/@GEOSgcm_App
   remote: ../GEOSgcm_App.git
-  tag: v1.3.16
+  branch: feature/esherman/gocart2g_develop_WilliamUpdate
   develop: develop

 UMD_Etc:
```
So if you want to make sure you get all the right bits for the CI, every unique branch needs to be named the same as the GOCART branch for CI mepo to see it.

mathomp4:
The final setup in CI was:
```
Checking out branch feature/esherman/gocart2g_develop_v2 in GEOSgcm
Checking out branch feature/esherman/gocart2g_develop_v2 in GEOSchem_GridComp
Checking out branch feature/esherman/gocart2g_develop_v2 in GOCART
Checking status...
GEOSgcm                | (b) feature/esherman/gocart2g_develop_v2
env                    | (t) v3.2.0 (DH)
cmake                  | (t) v3.3.7 (DH)
ecbuild                | (t) geos/v1.0.6 (DH)
NCEP_Shared            | (t) v1.1.1 (DH)
GMAO_Shared            | (t) v1.3.10 (DH)
MAPL                   | (t) v2.6.4 (DH)
FMS                    | (t) geos/2019.01.02+noaff.6 (DH)
GEOSgcm_GridComp       | (b) develop
FVdycoreCubed_GridComp | (t) v1.2.11 (DH)
fvdycore               | (t) geos/v1.1.5 (DH)
GEOSchem_GridComp      | (b) feature/esherman/gocart2g_develop_v2
GOCART                 | (b) feature/esherman/gocart2g_develop_v2
mom                    | (t) geos/5.1.0+1.1.1 (DH)
mom6                   | (t) geos/v2.0.1 (DH)
GEOSgcm_App            | (b) develop
UMD_Etc                | (t) v1.0.3 (DH)
CPLFCST_Etc            | (t) v1.0.1 (DH)
```

I suppose what I can do is to not require the GCM build for merges into `develop` for the time being?

tclune:
OK - I think we should park this PR then.    Hopefully Santha can give us a green light for another release of MAPL.  We can then get that into the GCM and the PR here should work.  Right?

ESMF/GOCART_GridComp/NI_GridComp/NI_GridCompMod.F90
gmao-esherman: GOCART Legacy does run NI. I  forgot to remove this print statement. I will remove this now.

mathomp4: Thanks. And there are a lot of them! I just commented on the first.

mathomp4:
> OK - I think we should park this PR then. Hopefully Santha can give us a green light for another release of MAPL. We can then get that into the GCM and the PR here should work. Right?

Yes. If we get a MAPL 2.6.x/2.7.0 with what's in `develop` *and* Scott takes it into `main` of GEOSgcm, then it would work.

ESMF/GOCART_GridComp/NI_GridComp/NI_GridCompMod.F90
gmao-esherman: Yes, very good catch! I have removed all the prints.

adarmenov (changes requested):

ESMF/GOCART2G_GridComp/DU2G_GridComp/DU2G_GridCompMod.F90
```diff
@@ -1122,5 +1104,158 @@ end subroutine mie_
...
+    else if ((wavelength .ge. 5.49e-7) .and. (wavelength .le. 5.51e-7)) then
+       mieTable_index = 2.
+    else if ((wavelength .ge. 6.69e-7) .and. (wavelength .le. 6.71e-7)) then
+       mieTable_index = 3.
+    else if ((wavelength .ge. 8.68e-7) .and. (wavelength .le. 8.71e-7)) then
+       mieTable_index = 4.
+    else
+       print*,trim(Iam),' : wavelength of ',wavelength,' is an invalid value.'
+       return
+    end if
```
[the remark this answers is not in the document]
tclune: Presuming this section is the same as the first, at the very least it should be encapsulated as a function.     I know @gmao-esherman and I discussed this at some point with @amdasilva but I no longer recall if the decision was to do it better later or if this is what he actually wanted.

ESMF/GOCART2G_GridComp/CA2G_GridComp/CA2G_GridCompMod.F90
```diff
@@ -1295,6 +1278,172 @@ end subroutine mie_
...
+    else if ((wavelength .ge. 5.49e-7) .and. (wavelength .le. 5.51e-7)) then
+       mieTable_index = 2.
+    else if ((wavelength .ge. 6.69e-7) .and. (wavelength .le. 6.71e-7)) then
+       mieTable_index = 3.
+    else if ((wavelength .ge. 8.68e-7) .and. (wavelength .le. 8.71e-7)) then
+       mieTable_index = 4.
+    else
+       print*,trim(Iam),' : wavelengths of ',wavelength,' is an invalid value.'
+       return
+    end if
```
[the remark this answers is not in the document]
gmao-esherman: There is a 'channels' variable within the mie table, but those values are set within the aerosol .rc file by the user and can change. I remember discussing this exact issue with hard-coding these values with @amdasilva and @tclune, but I don't fully remember why we settled on hard-coding the values. I think it's because the file used to construct the diagnostic mie tables is organized in this why, where the first index is for 450nm, the second index is for 550nm, etc.

I would be happy to implement a non-hard-coded solution if we can devise one!

ESMF/GOCART2G_GridComp/CA2G_GridComp/CA2G_GridCompMod.F90
```diff
@@ -1295,6 +1278,172 @@ end subroutine mie_
...
+
+    deallocate(q_4d, __STAT__)
+
+    RETURN_(ESMF_SUCCESS)
+
+  end subroutine monochromatic_aerosol_optics
+
+!---------------------------------------------------------------------------------------
+  subroutine get_mixR (state, rc)
```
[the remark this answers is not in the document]
gmao-esherman: Sure. This callback simply provides the sum of aerosol mixing ratios to be used in GAAS2G. Do you have a preferred name?

ESMF/GOCART2G_GridComp/DU2G_GridComp/DU2G_GridCompMod.F90
```diff
@@ -1122,5 +1104,158 @@ end subroutine mie_
...
+
+    deallocate(q_4d, __STAT__)
+
+    RETURN_(ESMF_SUCCESS)
+
+  end subroutine monochromatic_aerosol_optics
+
+!---------------------------------------------------------------------------------------
+  subroutine get_mixR (state, rc)
+
```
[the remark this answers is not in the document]
gmao-esherman: Good point. It's almost the same, but I probably could generalize these callbacks to work for all aerosols. I will give this a try.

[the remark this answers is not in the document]
gmao-esherman: I have generalized the subroutine get_mixR, renamed it get_mixRatioSum, and put it in Chem_AeroGeneric.F90. All aerosol gridded components now use get_mixRatioSum. get_mixR that used to live in each aerosol gridded component has been removed. This change is zero diff.

gmao-esherman:
Is there anything I can do to help move this PR along? I can't resolve the issue with hard-coded wavelength values without discussing with Arlindo.

ESMF/GOCART2G_GridComp/CA2G_GridComp/CA2G_GridCompMod.F90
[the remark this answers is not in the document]
adarmenov: How about get_total_mixing_ratio()

adarmenov (approved):

gmao-esherman:
It still look like the build test is failing in GAAS because it does not have MAPL develop.

gmao-esherman:
I was hoping to get these updates merged in before NOAA's new dust scheme was implemented. At this point I think it will be simpler to merge in these updates and the updated physics together in a single PR.

Issue #73, opened 2021-08-06
Clean up nm to m conversion in GOCART2G


mathomp4:
`wavelengths_diagmie` is still declared as `integer`: https://github.com/GEOS-ESM/GOCART/blob/58af7747eba91dc597c65485a5e4e781cb25deca/ESMF/GOCART2G_GridComp/GOCART2G_GridCompMod.F90#L113
which is inconsistent with `self%wavelengths_profile` and `self%wavelengths_vertint` (`real`).

I would still recommend including changes from PR #65 into this branch to address this issue and streamline the code. For example, conversion from `nm` to `m` should be done only once (see https://github.com/GEOS-ESM/GOCART/pull/65/files#diff-682724d291af0ad213c4e1b2ffa97156de4989543290b60bf362adfa98e3d8da):

https://github.com/GEOS-ESM/GOCART/blob/b9fa4c6c0ed46e43dae1a66aaa296a0d72ff839f/ESMF/GOCART2G_GridComp/GOCART2G_GridCompMod.F90#L156-L162

_Originally posted by @rmontuoro in https://github.com/GEOS-ESM/GOCART/pull/68#discussion_r684312869_

amdasilva:
This is less of an issue now since these calculations are now done at array level with the refactoring of the Mie calculations.

PR #65, opened 2021-07-30
Fix optical properties computation

rmontuoro:
This PR introduces changes addressing issue #64. 

`aerosol_monochromatic_optics_wavelength`s are now assumed to be provided in units of `nm` and internally converted to `m` for proper lookup using the diagnostic Mie table.

`aerosol_monochromatic_optics_wavelength` could be renamed as `aerosol_monochromatic_optics_wavelength_in_nm` for consistency with other configuration labels.

PR #74, opened 2021-08-10
Cleanup GOCART2G: nm-to-m and present

### LINKED ISSUE #72: Invalid Fortran in Process Library
@tclune - is this valid Fortran?
```
if (present(a) .and. associated(a)) then
```
Should this be:
```
if (present(a)) then
   if (associated(a)) then
```

_Originally posted by @rmontuoro in https://github.com/GEOS-ESM/GOCART/pull/68#discussion_r684301009_

mathomp4:
This PR is an attempt to close #72 and #73 which were brought up by @rmontuoro.

First, the nm-to-m issue (#73 ), this solution looks a little different to that in #65 because testing showed it needed to be. However, I believe it accomplishes the same thing.

Now for the more "fun" one, the issue with `present()`, #72 . The current code _might_ or _probably_ is a Fortran Standard violation (@tclune can say for certain), but it definitely "looks wrong":
```fortran
   if (present(aerindx) .and. associated(aerindx))  aerindx = 0.0  ! for now
```

The "fun" part is that the preferred way to do this would have been with a function using an assumed-rank array:
```fortran
      logical function present_and_associated(p) result(a)
         real, pointer, dimension(..), optional, intent(in) :: p
         a = .false.
         if (present(p)) then
            if (associated(p)) then
               a = .true.
            end if
         end if
      end function present_and_associated
```
but it turns out [the Intel compiler has a probable bug with this](https://community.intel.com/t5/Intel-Fortran-Compiler/Possible-bug-with-assumed-rank-array/m-p/1305251/emcs_t/S2h8ZW1haWx8dG9waWNfc3Vic2NyaXB0aW9ufEtTNjRISjM0R1NZQjk4fDEzMDUyNTF8U1VCU0NSSVBUSU9OU3xoSw#M157115), though GNU is happy. 

So, after quite a bit of test coding by @tclune and myself, it turns out the best way to handle this for now is with a macro:
```fortran
#define __PRESENT_AND_ASSOCIATED__(ptr,res) if(present(ptr)) then; res=associated(ptr); else; res=.false.; endif
```
This is uglier since we now have to make a logical variable for each test and then use that variable. So this:
```fortran
   if (present(aerindx) .and. associated(aerindx))  aerindx = 0.0  ! for now
```
becomes:
```fortran
   logical :: has_aerindx
...
   __PRESENT_AND_ASSOCIATED__(aerindx,has_aerindx)
   if ( has_aerindx )  aerindx = 0.0  ! for now
```
instead of a nice clean:
```fortran
   if ( present_and_associated(aerindx) )  aerindx = 0.0  ! for now
```

That said, @tclune might see some way to make this cleaner. 

Still, hopefully this all works. My testing shows it is zero-diff in the Ginoux and K14 cases compared to `develop` (aka v2.0.0-rc.1)

### DIFF
--- a/ESMF/GOCART2G_GridComp/CA2G_GridComp/CA2G_GridCompMod.F90
+++ b/ESMF/GOCART2G_GridComp/CA2G_GridComp/CA2G_GridCompMod.F90
@@ -550,10 +550,13 @@ contains
     call ESMF_ConfigGetAttribute (universal_cfg, self%diag_MieTable(instance)%channels, &
                                   label= "aerosol_monochromatic_optics_wavelength_in_nm_from_LUT:", __RC__)
 
+!   Convert input wavelengths from nm to m for internal use
+    self%diag_MieTable(instance)%channels = self%diag_MieTable(instance)%channels * 1.0e-9
+
     allocate (self%diag_MieTable(instance)%mie_aerosol, __STAT__)
     self%diag_MieTable(instance)%mie_aerosol = Chem_MieTableCreate (self%diag_MieTable(instance)%optics_file, __RC__ )
     call Chem_MieTableRead (self%diag_MieTable(instance)%mie_aerosol, self%diag_MieTable(instance)%nch, &
-                            self%diag_MieTable(instance)%channels*1.e-9, rc=status, nmom=self%diag_MieTable(instance)%nmom)
+                            self%diag_MieTable(instance)%channels, rc=status, nmom=self%diag_MieTable(instance)%nmom)
     VERIFY_(status)
 
 !   Finish creating AERO state
@@ -1031,7 +1034,7 @@ contains
     int_arr(:,:,:,2) = intPtr_philic
 
     call Aero_Compute_Diags (mie_table=self%diag_MieTable(self%instance), km=self%km, klid=self%klid, nbegin=1, nbins=2, &
-                             channels=self%diag_MieTable(self%instance)%channels*1.0e-9, &
+                             channels=self%diag_MieTable(self%instance)%channels, &
                              wavelengths_profile=self%wavelengths_profile, &
                              wavelengths_vertint=self%wavelengths_vertint, aerosol=int_arr, grav=MAPL_GRAV, &
                              tmpu=t, rhoa=airdens, rh=rh2, u=u, v=v, delp=delp, ple=ple, tropp=tropp, &
--- a/ESMF/GOCART2G_GridComp/CA2G_GridComp/CA2G_GridCompMod.F90
+++ b/ESMF/GOCART2G_GridComp/CA2G_GridComp/CA2G_GridCompMod.F90
@@ -1032,8 +1032,8 @@ contains
 
     call Aero_Compute_Diags (mie_table=self%diag_MieTable(self%instance), km=self%km, klid=self%klid, nbegin=1, nbins=2, &
                              channels=self%diag_MieTable(self%instance)%channels*1.0e-9, &
-                             wavelengths_profile=self%wavelengths_profile*1.0e-9, &
-                             wavelengths_vertint=self%wavelengths_vertint*1.0e-9, aerosol=int_arr, grav=MAPL_GRAV, &
+                             wavelengths_profile=self%wavelengths_profile, &
+                             wavelengths_vertint=self%wavelengths_vertint, aerosol=int_arr, grav=MAPL_GRAV, &
                              tmpu=t, rhoa=airdens, rh=rh2, u=u, v=v, delp=delp, ple=ple, tropp=tropp, &
                              sfcmass=CASMASS, colmass=CACMASS, mass=CAMASS,&
                              exttau=CAEXTTAU,stexttau=CASTEXTTAU, scatau=CASCATAU, stscatau=CASTSCATAU,&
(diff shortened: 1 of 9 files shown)


ESMF/GOCART2G_GridComp/CA2G_GridComp/CA2G_GridCompMod.F90
```diff
@@ -550,10 +550,13 @@ subroutine Initialize (GC, import, export, clock, RC)
     call ESMF_ConfigGetAttribute (universal_cfg, self%diag_MieTable(instance)%channels, &
                                   label= "aerosol_monochromatic_optics_wavelength_in_nm_from_LUT:", __RC__)
 
+!   Convert input wavelengths from nm to m for internal use
+    self%diag_MieTable(instance)%channels = self%diag_MieTable(instance)%channels * 1.0e-9
```
rmontuoro, line 554: This is confusing to me. I have a hard time seeing the need of performing this `nm`-to-`m` conversion multiple times (in different components) when this could be done once in the GOCART 2G gridded component and stored in `universal_cfg`, to be retrieved by all other components as needed in the proper units.

ESMF/GOCART2G_GridComp/GA_Environment/GA_EnvironmentMod.F90
```diff
@@ -72,6 +72,11 @@ subroutine load_from_config(self, cfg, universal_cfg, rc)
        call ESMF_ConfigGetAttribute (universal_cfg, self%wavelengths_vertint, &
                                      label='wavelengths_for_vertically_integrated_aop_in_nm:', __RC__)
 
+       !   Convert input wavelengths from nm to m for internal use
+
+       self%wavelengths_profile = 1.0e-9 * self%wavelengths_profile
```
rmontuoro, line 77: See previous comment.

rmontuoro (changes requested):
Looks fine but I'd like to understand more about the unit conversion, since the code could potentially be further streamlined.

ESMF/GOCART2G_GridComp/CA2G_GridComp/CA2G_GridCompMod.F90
mathomp4: I think the main issue is that we must do the conversion after the get from the resource file. Since above this we are getting `aerosol_monochromatic_optics_wavelength_in_nm_from_LUT` from `universal_cfg` into `self%diag_MieTable(instance)%channels`, we convert to meters after this. There is no point before this where `self%diag_MieTable(instance)%channels` is filled.

ESMF/GOCART2G_GridComp/GA_Environment/GA_EnvironmentMod.F90
mathomp4: Again as above, here we are filling `self%wavelengths_vertint` from `wavelengths_for_vertically_integrated_aop_in_nm`, so we convert to meters after that. This one is the `self%wavelengths_vertint` that each "subcomponent" sees. As each one (DU, CA, etc.) calls `load_from_config` to fill their respective states.

@tclune might have more information.

ESMF/GOCART2G_GridComp/CA2G_GridComp/CA2G_GridCompMod.F90
rmontuoro: @mathomp4 - I understand your point and generally agree with it. However, in this case, optical properties are not read from a configuration file associated with `universal_cfg`, but rather _set_ in `universal_cfg` by the GOCART 2G gridded component (`universal_cfg` is used similarly to an internal state):

https://github.com/GEOS-ESM/GOCART/blob/e3682659c00d317652d4743963a191a5e63ae67f/ESMF/GOCART2G_GridComp/GOCART2G_GridCompMod.F90#L156-L160

This is why I'm suggesting that we could perform the conversion only once before setting the above values in `universal_cfg`. I also understand that setting wavelengths in `m` for internal use will contrast with the updated configuration labels in `universal_cfg`, which explicitly include `nm` in their name.

mathomp4: Maybe, but the label ***explicitly*** says `in_nm` and I cannot go against that. If a scientist wishes to make that change, okay, but for now, I obey the label.

Plus, what if a scientist in the future pulls from the config expecting it to be in `nm` and they get `m`?

Can you show @tclune or I a profile of UFS-Aerosols showing this multiplication is a tall-pole in your runs? I can't imagine it is that expensive. I'll try and dump the assembly code tomorrow to see, but I can't imagine it is that many instructions, especially seeing as we do this in the initialize method. Does UFS call the initialize method multiple times such that this is a high cost?

rmontuoro: @mathomp4 - I do understand all of your points and I'm totally fine with merging this PR as is. My suggestion was mostly driven by the desire to simplify the code base, which may help prevent bugs in the future.

rmontuoro (approved):

tclune:
My prediction is that it does not even have a cost.   There are so many other non FP operations lying around this bit of code, that the compiler can almost certainly bury these (scalar) conversions within some superscalar instruction that is accomplishing something else.      But regardless, compared to the inefficiency of retrieving the value from the resource in the first place such operations are inconsequential.   

Perhaps I've missed a bit of the back and forth, but my understanding is that the conversions would only happen once when the quantity in question is first retrieved from the resource file.   So it does not even contribute during the run phase.

PR #77, opened 2021-09-07
Remove Legacy GOCART code

mathomp4:
I realized I never made a PR for this. This removes the old legacy code from GOCART. It is technically non-zero-diff for GOCART2G but only because of MAPL grid handling (the ExtData race condition, see @bena-nasa )


mathomp4:
I suppose I should ping @rmontuoro as well about this. I don't think it should matter to UFS.

adarmenov (approved):

rmontuoro (approved):
Looks good. Thank you, @mathomp4, for including me.

mathomp4:
I believe #81 by @amdasilva supersedes this PR.

PR #85, opened 2021-12-06
Updated dust tuning parameters

amdasilva:
This slight retuning of the dust emissions is needed since we are back using the Ginoux scheme. This is needed for a new GOCART release as we continue valiation for inclusion of GOCART-2G into GEOSgcm.


adarmenov (dismissed):

amdasilva:
This should be merged into develop, not main. My mistake.

mathomp4 (approved):
Approve after merge

PR #90, opened 2022-01-19
Clarified keyword usage in call to Mie calculations

pcolarco:
Added argument names for Mie calculation call for clarification for optional keywords and added to CHANGELOG; 0 diff

PR #94, opened 2022-02-09
Edits to make GNU work with GOCART2G

mathomp4:
Work by @aoloso has found that the GNU compiler does not work with GOCART2G. Part of the issue was in Chem::Achem and is detailed in https://github.com/GEOS-ESM/GEOSchem_GridComp/pull/161

TL;DR This PR with https://github.com/GEOS-ESM/GEOSchem_GridComp/pull/161 allows GNU to work and is zero-diff!

---

Here in GOCART we need another fix. The Fortran construct consisting of, for example:
```fortran
real, optional, pointer, dimension(:,:,:), intent(inout)   :: exttau    ! ext. AOT at 550 nm
...
if (present(exttau) .and. associated(exttau)) then
```
is a Fortran Gotcha. The reason is that the compiler is free to evaluate that if-test by testing the `associate()` first. And if the pointer is not `associated()`, then GNU (at least) is not happy as associated on a non-present pointer is...ill-defined at best?

But, thanks to @tclune reaching out to Fortran Committee, an interesting solution was presented. This construct:
```fortran
real, optional, dimension(:,:,:), intent(inout)   :: exttau    ! ext. AOT at 550 nm
...
if (present(exttau)) then
```
is allowed and legal. Per Malcolm Cohen (of NAG and [that book on your shelf](https://global.oup.com/academic/product/modern-fortran-explained-9780198811886?cc=us&lang=en&)):

> disassociated pointers are “absent” to nonpointer optionals

So this means we can remove the `pointer` declaration and then use the fact that the Standard says a null'd pointer passed as a nonpointer optional is not-present!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+## Fixed
+
+- Fixes to allow GNU to run GOCART2G
+  - Changes to `if (present(foo)) .and. associated(foo))` construct which is ambiguous Fortarn
+
 ## [2.0.2] - 2021-1-07
 
 ### Changed
@@ -12,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Update CI to Baselibs 6.2.8 and GCC 11.2
 - Updated dust emissions tuning to be consistent with current Ginoux default.
 
-## [2.0.1] - 2021-10-20 
+## [2.0.1] - 2021-10-20
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


mathomp4:
Note that:

https://github.com/GEOS-ESM/GEOSchem_GridComp/pull/161

was pulled into the [`feature/#328/geos-v10.19.4/gocart2g`](https://github.com/GEOS-ESM/GEOSchem_GridComp/tree/feature/%23328/geos-v10.19.4/gocart2g) branch (it was a bunch of nulls).

Hopefully we can get a v2.0.3 with this fix (and #90 by @pcolarco as well) before @sdrabenh finalizes a GOCART2G GEOSgcm

mathomp4:
Blocking as I might have an issue.

mathomp4:
Okay. I think I'll need to consult with @pcolarco or @christophkeller or someone who knows aerosols. My fixes here (which I *think* are good valid Fortran and I'll consult with @tclune about that) are zero-diff for state and are zero-diff for *ALL* AGCM HISTORY output except:

- DUEXTT25001
- DUEXTTFM001
- DUSCAT25001
- DUSCATFM001
- SSEXTT25001
- SSEXTTFM001
- SSSCAT25001
- SSSCATFM001

These change, and the change is not insignificant in the few I've looked at.

mathomp4:
Thanks to @pcolarco I think I have the right code. It's now all zero-diff even history!

It was my fault. I'll consult with @tclune on the best way to code it up and then push.

pcolarco:
> Thanks to @pcolarco I think I have the right code. It's now all zero-diff even history!
> 
> It was my fault. I'll consult with @tclune on the best way to code it up and then push.

Sounds good!

tclune (approved):

PR #116, opened 2022-03-10
added AMIP.20C emission scenario

sdrabenh:


mathomp4:
I am moving this to draft in favor of #120. Something wonky happened here and the CI is confused.

PR #119, opened 2022-03-12
use elemental for interpolation

weiyuan-jiang:
This is just for comparison.

### DIFF
--- a/Process_Library/GOCART2G_MieMod.F90
+++ b/Process_Library/GOCART2G_MieMod.F90
@@ -39,9 +39,9 @@ module GOCART2G_MieMod
 ! --------
    integer, parameter :: NRH_BINS = 991
 
-   type :: RH_Table
+   type :: RH_Mie
       real, allocatable :: rh(:)
-   end type RH_Table
+   end type RH_Mie
 
    type GOCART2G_Mie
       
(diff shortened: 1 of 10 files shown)


Process_Library/GOCART2G_MieMod.F90
```diff
@@ -326,73 +330,88 @@ type(GOCART2G_Mie) function GOCART2G_MieCreate ( rcfile, wavelengths, nmom, rc )
...
+        this%gf(j)%rh   = gf_table(:,j)
+        !Wet particle density [kg m-3]
+        this%rhop(j)%rh = rhop_table(:,j)
+        !Dry particle density [kg m-3]
+        this%rhod(j)%rh = rhod_table(:,j)
+        !Volume [m3 kg-1]
+        this%vol(j)%rh  = vol_table(:,j)
+        !Area [m2 kg-1]
+        this%area(j)%rh = area_table(:,j)
+      enddo
```
tclune, line 358: There should be a procedure that fully allocates each component of the derived type.  (Might need 2 or 3 such to handle the oddballs like pmom and rEff.)

Process_Library/GOCART2G_MieMod.F90
```diff
@@ -326,73 +330,88 @@ type(GOCART2G_Mie) function GOCART2G_MieCreate ( rcfile, wavelengths, nmom, rc )
...
+          allocate (this%bext(n,j)%rh(this%nrh), __NF_STAT__)
+          allocate (this%bsca(n,j)%rh(this%nrh), __NF_STAT__)
+          allocate (this%bbck(n,j)%rh(this%nrh), __NF_STAT__)
+          allocate (this%g(n,j)%rh(this%nrh), __NF_STAT__)
+          allocate (this%refr(n,j)%rh(this%nrh), __NF_STAT__)
+          allocate (this%refi(n,j)%rh(this%nrh), __NF_STAT__)
+          do i = 1, this%nrh
+            call polint(channels_table,bext_table(:,i,j),nch_table, &
+                     this%wavelengths(n),this%bext(n,j)%rh(i),yerr)
+            call polint(channels_table,bsca_table(:,i,j),nch_table, &
```
tclune, line 384: How ugly would it be to move the `i` index to inside the polint procedure?

Process_Library/MieQuery.H
```diff
@@ -87,76 +87,73 @@
...
 #endif
 
-    rh_ = reshape(min(max(rh,0.),0.99),[size(rh)])    ! no (super) saturation 
-    qrh = max(1,nint((rh_+0.001)*1000.))              ! quantized RH
-    arh = this%rha(qrh)                               ! interpolation weight in table
-    irh = this%rhi(qrh)                               ! index in table for each RH vcalue
+    rh_ = min(max(rh,0.),0.99)                                  ! no (super) saturation 
+    qrh = reshape(max(1,nint((rh_+0.001)*1000.)), [size(rh_)])  ! quantized RH
+    arh = reshape(this%rha(qrh), shape(rh_))                    ! interpolation weight in table
+    irh = reshape(this%rhi(qrh), shape(rh_))                    ! index in table for each RH vcalue
```
tclune, line 93: The point of this approach was supposed to be that we could skip the `reshape()` steps.

Process_Library/GOCART2G_MieMod.F90
```diff
@@ -537,26 +556,35 @@ integer function getChannel(this, wavelength, rc) result (ch)
...
 
+  elemental function interp(this,table, irh, arh)
+    class (GOCART2G_Mie), intent(in) :: this
+    type(RH_Table), intent(in) :: table
+    integer, intent(in) :: irh
+    real, intent(in) :: arh
+    real :: interp
+    interp = sum(table%rh(irh:irh+1) * [(1-arh),arh])
+  end function interp
+
```
tclune, line 589: This should not be a type-bound procedure.   It does not use `this` inside.

Process_Library/GOCART2G_MieMod.F90
```diff
@@ -86,7 +90,7 @@ module GOCART2G_MieMod
                             QueryByChannel_3d
       procedure :: getChannel
       procedure :: getWavelength
-      
+      procedure :: interp
```
tclune, line 93: This should not be a type-bound procedure.   See comment at end.

Process_Library/MieQuery.H
```diff
@@ -87,76 +87,73 @@
...
 
     if (present(pmom)) then
-       allocate(pmom_(product(shape(irh)),size(this%pmom,4),size(this%pmom,5)))
-       do j = 1, size(this%pmom,5)
-          do i = 1, size(this%pmom,4)
-             pmom_(:,i,j)  =  this%pmom(irh   ,channel,bin,i,j) * (1.-arh) &
-                            + this%pmom(irh+1 ,channel,bin,i,j) * arh
+       do j = 1, size(this%pmom,4)
+          do i = 1, size(this%pmom,3)
+             pmom(__DIMS__,i,j) = this%interp(this%pmom(channel,bin,i,j),irh,arh)
```
tclune, line 154: One reason that I think Arlindo may like this version is that it only needs a single `interp` rather than 3 similar macros.

But he also still may want a macro to simply the expressions above.   E.g.,
```
#define __INTERP__(name, irh, arh)   interp(this% name (channel, bin), irh, arh)
```
(But would still need another macro for pmom, so maybe not.

Process_Library/GOCART2G_MieMod.F90
weiyuan-jiang: it is easy. Done

Process_Library/MieQuery.H
weiyuan-jiang: It is hard to avoid. If we put it inside subroutine interp, it will have to be calculated every time

Process_Library/GOCART2G_MieMod.F90
weiyuan-jiang: Done

Process_Library/MieQuery.H
weiyuan-jiang: There is no macro __INTERP__ for this version and this line is simple enough. So I prefer not to have one

weiyuan-jiang:
it is not necessary anymore. closed

PR #123, opened 2022-03-14
Update CircleCI CI to use Orb. Update components

mathomp4:
This PR updates the CI here to use the more modern circleci-tools orb for reusable configs. This also has the benefits of builds with both gfortran and intel!

It also updates the `components.yaml` to better match GOCART2G GEOSgcm.

I've also added a `.editorconfig` file to match the styles used in MAPL
- 2 space indents for CMake and YAML
- 4 space indents for Python (per PEP 8)

Note that for most legacy editors (emacs, vim, etc.), [EditorConfig](https://editorconfig.org/) needs a plugin to be enabled, so for most people this is a no-op, but if you have it on, it helps enforce a style for a few files.


tclune (approved):
CI changes ok.

adarmenov (approved):
Changes in HEMCO are harmless. Approved.

tclune (approved):

amdasilva (approved):
Looks fine...

PR #124, opened 2022-03-18
Update CODEOWNERS and update components to match GEOSgcm v10.22.1

mathomp4:
This PR updates the `CODEOWNERS` file so that PR approvals here are less onerous. Only the CI and weird CCPP/UFS directories will need approval from non-aerosol team members.

NOTE: I think this will not truly take effect until this change is in `main` due to the way GitHub works. (Only the `CODEOWNERS` on `main` matters to GitHub. Annoying, but it's how they do things.)

---

ETA: I've also updated `components.yaml` to match that of GEOSgcm v10.22.1. The main thing is to get MAPL 2.19.0 which is needed for the new AGC code that is in `develop`.

PR #126, opened 2022-03-21
Updates for Spack Support, Round 2

mathomp4:
Due to insufficient grepping on my part, some remnants of the
"not-canonical" CMake still remain. Apparently these were not necessary
changes for Spack, but they are changes that unify the CMake style of
GEOS to be more like "correct" CMake.

In this go-around, it is mainly finding `INC_ESMF` which I apparently
forgot to grep for last time (I was focused on NetCDF), and some
non-canonical OpenMP calls.

See https://github.com/GEOS-ESM/GEOSgcm/issues/387 for the meta-issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,9 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - This matches the styles currently used in MAPL (2 space indents in CMake and yaml, 4 spaces for Python)
 
 ### Changed
+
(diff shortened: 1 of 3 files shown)


tclune (approved):
cmake changes ok

mathomp4:
Whoops. Since #124 wasn't in yet, the CI here fails. So, I updated the `components.yaml` in this PR as well so that CI can pass.

adarmenov (approved):

amdasilva (approved):
Looks fine.

Issue #129, opened 2022-03-24
GOCART develop doesn't currently run with AMIP (at least)


mathomp4:
This might be something @amdasilva knows about, but just in case...

When I clone GEOSgcm `main` and then checkout the `develop` branch of GOCART, an AMIP run fails:

```
 Character Resource Parameter: DU_INTERNAL_RESTART_FILE:du_internal_rst
 Bootstrapping du_internal_rst
 Character Resource Parameter: SS_INTERNAL_RESTART_FILE:ss_internal_rst
 Bootstrapping ss_internal_rst
 Character Resource Parameter: CA.oc_INTERNAL_RESTART_FILE:caoc_internal_rst
 Bootstrapping caoc_internal_rst
pe=00091 FAIL at line=00461    CA2G_GridCompMod.F90                     <status=41>
pe=00091 FAIL at line=01799    MAPL_Generic.F90                         <status=41>
```

It seems to die on this line:
https://github.com/GEOS-ESM/GOCART/blob/9737da633cfc937969799be6cabe2cbd496aeb14/ESMF/GOCART2G_GridComp/CA2G_GridComp/CA2G_GridCompMod.F90#L461

~Perhaps `HISTORY.rc` needs a change for GOCART `develop`? Not sure.~ Nope. Does not involve history.

amdasilva:
Can you be more specific?  If this is HISTORY related, it has nothing to do with AMIP.

mathomp4:
@amdasilva What I did was checkout `develop`, build, and then try to run an AMIP-emissions run for a day (the way I normally run things). And it crashed. I'm not sure why. I was just sort of idly speculating. 

But you are right, this does not involve History. I turned off all the collections, and the model still crashed at the same place.

weiyuan-jiang:
The recent changes don't seem to involve those errors. Need further investigation...

Issue #93, opened 2022-02-08
Correct handling of state vars in multiple instances of component


tclune:
From @amdasilva 
>The other issue is the variable names in the multiple-instances of a component, CA, in particular.
```
CAMASS*    | kg kg-1    | xyz | C   |       | Carbonaceous Aerosol Mass Mixing Ratio                
CACONC*    | kg m-3     | xyz | C   |       | Carbonaceous Aerosol Mass Concentration               
CAEXTCOEF* | m-1        | xyz | C   | size(self%wavelengths_profile) | Carbonaceous Aerosol Extinction Coefficient 
CASCACOEF* | m-1        | xyz | C   | size(self%wavelengths_profile) | Carbonaceous Aerosol Scattering Coefficient
```
 
>In the case that the instance is called CA.oc, the first variable becomes CAMASSOC, while it should have been OCMASS for backward compatibility. Then long name is now the same for OC, BC and BR.  Should not have been done this way, but instead

``` 
*MASS    | kg kg-1    | xyz | C   |       | * Carbonaceous Aerosol Mass Mixing Ratio                
*CONC    | kg m-3     | xyz | C   |       | * Carbonaceous Aerosol Mass Concentration               
*EXTCOEF | m-1        | xyz | C   | size(self%wavelengths_profile) | * Carbonaceous Aerosol Extinction Coefficient 
*SCACOEF | m-1        | xyz | C   | size(self%wavelengths_profile) | * Carbonaceous Aerosol Scattering Coefficient
```

>Where * gets replaced by OC, BC, BR. Also he has code like this that is not used for anything. 

```f90
    if (comp_name(1:5) == 'CA.oc') then
       GCsuffix = 'OC'
    else if (comp_name(1:5) == 'CA.bc') then
       GCsuffix = 'BC'
    else if (comp_name(1:5) == 'CA.br') then
       GCsuffix = 'BR'
    end if
```

>As it turns out, the variables are unnecessarily different.  Also, no provisions for multiple instances in the other components.

@tclune will make any needed corrections to the MAPL ACG for this.    This ticket is solely about correcting the usage in GOCART gridcomps when those fixes are available.

weiyuan-jiang:
Where is the original issue ?

amdasilva:
@tclune Have the updates above been implemented in ACG?

tclune:
[...]

PR #135, opened 2022-03-25
rewind and fix bug in MieQuery.H (use a shape of a not present variable)

weiyuan-jiang:
I would like rewind Issue #93 at this moment. There are many hard-coded variable names that need to be changed. This PR addresses the issue #129 . Also a bug in MieQuery.H is fixed.  We should have a separate PR to address #93


weiyuan-jiang:
closed. PR  #138 can address this issue

Issue #128, opened 2022-03-23
Split process library using Fortran submodules


tclune:
Submodules will provide for less contention as individual procedures are modified while preserving the namespace protections of a module.

If there are natural groupings among the routines multiple submodules would be appropriate.  Otherwise one large one.   

Naming of submodules and associated file is in a bit of flux.  With existing GOCART2G conventions, I recommend module has `SMod` at the end and the file also has `...Smod.F90`.  (Once we have something working, we can let @amdasilva suggest tweaks to the style.)

Mentioning @rmontuoro for the suggestion to split into multiple files.

weiyuan-jiang:
Before I split the file, would anybody help me to divide the subroutines into different groups roughly?   @tclune @rmontuoro

amdasilva:
I am still not sold on this submodule thing.  Too much busy work for little benefit.

tclune:
@weiyuan-jiang  I spoke with @amdasilva late this afternoon.   Rather than using submodules he recommends that we simply put each procedure into a separate source file with the same name.  (I.e.,  SUBROUTINE XYZ  goes into a file called XYZ.F90).   Then just use `#include ...` to bring them into the actual module.

The extra wrinkle is to change the module from default `PRIVATE` to default `PUBLIC`.   This will reduce the amount of duplication when adding new procedures in the future. You'll then need to make a handful of the entities `PRIVATE` explicitly.   E.g., the derived types, parameters, and other things brought in by USE statements.  (I.e., the public/private status for any given entity does not change; just the default.)

amdasilva:
We need a meeting to discuss this.

PR #137, opened 2022-03-29
split GOCART2G_Process.F90.

weiyuan-jiang:
Each public subroutine now is in a separate file. It addresses issue #128


amdasilva (changes requested):
Rejecting PR. While splitting of Process Library is desirable, we need to make the split taking in consideration functionality, not a simple 1 file per subroutine split. Let's schedule a meeting to discuss best split.

PR #150, opened 2022-05-23
new files for ExtData2G

bena-nasa:
@amdasilva  This brings in the new yaml files for ExtData2G so those will be available when the switch is thrown to start using them. This is zero diff as I'm not getting rid of the old files yet (and ExtData2G is zero diff in my testing anyway). Note this PR can't be tested **BECAUSE THE DEVELOP BRANCH IS NOT WORKING!**  (as of the time I created this PR, I was testing these new files with main but just want to get these in somewhere, note that there's been a PR sitting for over a month from @weiyuan-jiang  that fixes this,  but will require updating the GEOSgcm_Apps repo too)
If no one is doing to take the time to get develop working, please let me know and we will have to make a new release with these off of main but these **MUST** get into a release.


adarmenov (approved):

amdasilva (approved):
Mods seems reasonable but this has not been thoroughly tested yet.  Let's keep pour fingers crossed.

PR #152, opened 2022-05-23
Comment out anthropogenic emissions through ExtData

acollow:
This is part of the effort to implement scaling factors for diurnal and day of the week scaling factors through HEMCO and is contingent upon changes to the HEMCO and GEOSchem_GridComp repositories.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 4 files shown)


bena-nasa:
@acollow why you are commenting these out. If you can safely comment these out, then they aren't were not needed in the first place and they were obsolete or unused code. You should delete the line if it is not needed, otherwise it is confusing. Also, shouldn't this be going to the development branch of GOCART. The main is for hotfixes only unless @amdasilva has a different Git policy other than standard Git flow.

acollow:
Thanks for the heads up @bena-nasa! Would you recommend removing those lines and submitting a new PR to develop? The lines that were commented out were added to HEMCOgocart2g_ExtData.rc. This was my first PR so apologies for all the confusion!

PR #153, opened 2022-05-23
Feature/acollow/hemcoscaling

acollow:
Remove lines for anthropogenic emissions of OC, BC, SO2, and NH3 from their respective ExtData RC files as these emissions are pulled in through HEMCO.


bena-nasa:
@acollow I think you misunderstood me. In the previous PR, you commented the line in the ExtData.rc file out, now you have left it in but set the file template to /dev/null

I was suggesting that you delete the entire line, this PR is doing something entirely different than your last. Please call me on teams and let me know what you are trying to do if need clarification

acollow:
@bena-nasa, the lines for the anthropogenic emissions no longer pulled in through GOCART/ExtData are now removed.

Issue #148, opened 2022-05-19
du_src is null when scheme is not 'k14' or 'ginoux'


weiyuan-jiang:
It would crash here  https://github.com/GEOS-ESM/GOCART/blob/821ef40859d6155c8594668e6261c97c49a935bc/ESMF/GOCART2G_GridComp/DU2G_GridComp/DU2G_GridCompMod.F90#L745

Issue #127, opened 2022-03-21
GOCART run failed in ufs-weather-model with gnu compiler 


junwang-noaa:
The ufs-weather-model coupled test with GOCART failed with gnu compiler gnu/10.1.0 (on Cheyenne) and gnu/9.2.0 (on Hera). The compile jobs actually ran successfully. The test failed at run time with the same error message:

Backtrace for this error:
Program received signal SIGSEGV: Segmentation fault - invalid memory reference.
...
117: #1  0x12f5195 in modeldatainitialize
117:  at /scratch1/NCEPDEV/stmp4/Minsuk.Ji/ufs-aerosol-gnu-issue/GOCART/ESMF/UFS/Aerosol_Cap.F90:321  

An issue was created on ufs-weather-model at https://github.com/ufs-community/ufs-weather-model/issues/1115.

May I ask if there is any suggestion on fixing the issue. Thanks

tclune:
@junwang-noaa Can you post the version/tag of GOCART to this ticket so that we can look at the specific line that is referenced?   And please remind me if you have any ability to test with a more recent gfortran.

tclune:
Oh and please post the tag/version of MAPL here as well.  (Now that I've seen the likely line in the Aerosol cap.

tclune:
@junwang-noaa @rmontuoro   I strongly suspect that someone at my end will need to build and run your system to replicate the problem to make any progress.   The failure message in the original post really does not give us much to go on.  

My bet would be that it is a compiler bug and that a workaround is probably possible.   But if a newer GNU compiler resolves the issue, that would be the preferred solution.   At the very least, I suggest 10.3 if not 11.2.

junwang-noaa:
@tclune Thank you very much for looking into it!  In ufs-weather-model, we are using GOCART 2.0.5:
https://github.com/GEOS-ESM/GOCART/tree/1ff638919fad66e79120eef07daf835c6fee12a4

The line giving error is at:
https://github.com/GEOS-ESM/GOCART/blob/1ff638919fad66e79120eef07daf835c6fee12a4/ESMF/UFS/Aerosol_Cap.F90#L321

and MAPL: 2.11.0:
https://github.com/ufs-community/ufs-weather-model/blob/develop/modulefiles/ufs_common#L23

junwang-noaa:
Thanks, Tom. I will check if we have gnu 10.3 or 11.2 available.

jkbk2004:
[...]

PR #157, opened 2022-05-27
Bugfix/wjiang/gnu pointer allocation

weiyuan-jiang:
This PR is addressing issue #148 and issue #127


ESMF/UFS/Aerosol_Cap.F90
```diff
@@ -311,14 +311,16 @@ subroutine ModelDataInitialize(model, rc)
     end if
 
     ! startup MAPL
-    allocate(is % wrap % maplCap, stat=stat)
+    allocate(is % wrap % maplCap, stat=stat, source = MAPL_Cap("MAPL Driver", aerosol_routine_SS, cap_options=maplCapOptions, rc=rc))
```
tclune, line 314: I don't think `stat` is the right variable to check here.   Should be the `rc` code from the previous call.     Please also check the other calls to the log error routines.

weiyuan-jiang: After this call, both stat and rc are checked.  It should be fine

tclune (approved):
"CMake" changes ok.

vbuchard (approved):
Looks good to me.

PR #158, opened 2022-05-27
Fixes #156

### LINKED ISSUE #156: GOCART develop crashing with out of bounds error after updates
After the PR merged into develop today, when I tried using the gocart develop branch, it crashed with this out of bounds error when using the "debug" build:
```
forrtl: severe (408): fort: (3): Subscript #3 of the array DPFLI has value 0 which is less than the lower bound of 1
Image              PC                Routine            Line        Source
GEOSgcm.x          000000000F493DCF  Unknown               Unknown  Unknown
GEOSgcm.x          000000000146F764  gocart2g_process_        6302  GOCART2G_Process.F90
GEOSgcm.x          0000000001158F67  su2g_gridcompmod_        1078  SU2G_GridCompMod.F90
libesmf.so         00002AAAAB40A9E4  _ZN5ESMCI6FTable1     Unknown  Unknown
```
The odd thing is that if you examine the code, it seems like it should be protected from getting there
```
6299 if(LH .lt. 1) continue
6300
6301 do k = LH, km
6302 qls(k) = dpfli(i,j,k)/pdog(i,j,k)*rhoa(i,j,k)
6303 end do
```
I tried changing the continue to cycle, thinking that maybe it was that but same behaviour. Will continue investigating.

bena-nasa:
This fixes issue #156 
There continues that should have been cycles in the GOCART2G process library. This was causing the loop not to break when it should have and an out of bounds error to occur.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   radiative forcing calculation; zero diff change
 - Simplified loading of radiation MieTables.
 
+
 ## [2.0.7] - 2021-04-29
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


vbuchard (approved):
Looks good to me.

Issue #161, opened 2022-06-09
GOCART components and CI need to be updated


mathomp4:
We've noticed some issues with out-of-date CI in other repos, and GOCART CI is still using the 0.13 orb. This should be updated to v1.

Also, the components are now slightly out of date with the current GEOSgcm, so I can bring that up to date too.

PR #168, opened 2022-07-22
Feature/vbuchard/pm25 fix

vbuchard:
Fixed a typo in the calculation of surface PM2.5 (check if nitrate is active for not double counting ammonium when adding sulfate)


mathomp4:
@vbuchard for this to pass the build tests, you'll need to merge in #161 and then update this branch so that it gets the new CI. 

If you like, I could use superpowers to pull in #161 (since I think both Tom and Arlindo might be on leave at the moment?)

amdasilva (approved):
Looks reasonable.

PR #169, opened 2022-07-29
Export SCACOEF fields; add Is_G2G_Instance_Running

mmanyin:
In order for siblings of GOCART to access exports of GOCART2G aerosol species (via MAPL AddConnectivity in GOESchem_GridComp), the GOCART2G GridComp must export them.  This PR takes a first crack at exporting SCACOEF from BC, OC, BR, DU, SU, SS and NI.  It also adds a utility routine so the Chem GridComp can determine whether certain aerosol instances are running.

This PR is zero diff.


amdasilva (changes requested):
Providing a  routine for returning the details of what is running in GOCART-2G  appears to be an architectural violation. No such detail should be exposed. Rejecting this PR as is. Please come talk to me.

mmanyin:
We will re-factor to use a Total Scattering Coefficient export from GOCART2G.  This will address the architectual concerns, and preserve the encapsulation of G2G.  Closing this PR.

PR #170, opened 2022-08-08
Feature/mmanyin/delete code artifact

mmanyin:
This is zero diff.
Remove the declaration of the routine "disable_convection", which is no longer used.


amdasilva (approved):
This is good for now. We will need to revisit this once we disable wet removal in MOIST.

amdasilva (approved):
Approved.

Issue #171, opened 2022-08-16
Problem in the nitrogen component when number of DU or SS bins are modified


bbakernoaa:
After running an experiment changing the number of dust bins I noticed that the output AOD_NI is all NaNs.  This is due to the `NIheterogenousChem` call being hard coded for the number of dust and seasalt bins, which can be seen [here](https://github.com/GEOS-ESM/GOCART/blob/bb4eb2c42c3b0b290b6113dc4c628400e526ac6a/ESMF/GOCART2G_GridComp/NI2G_GridComp/NI2G_GridCompMod.F90#L807-L813)

```fortran
 
call NIheterogenousChem (NIHT, self%xhno3, MAPL_UNDEF, MAPL_AVOGAD, MAPL_AIRMW, &
                             MAPL_PI, MAPL_RUNIV/1000., airdens, t, rh2, delp, DU, &
                             SS, self%rmedDU*1.e-6, self%rmedSS*1.e-6, &
                             self%fnumDU, self%fnumSS, 5, 5, self%km, self%klid, &
                             self%cdt, MAPL_GRAV, fMassHNO3, fMassNO3, NO3an1, NO3an2, & 
                             NO3an3, HNO3CONC, HNO3SMASS,  HNO3CMASS, __RC__)
```

Suggested fix 
```fortran
 
call NIheterogenousChem (NIHT, self%xhno3, MAPL_UNDEF, MAPL_AVOGAD, MAPL_AIRMW, &
                             MAPL_PI, MAPL_RUNIV/1000., airdens, t, rh2, delp, DU, &
                             SS, self%rmedDU*1.e-6, self%rmedSS*1.e-6, &
                             self%fnumDU, self%fnumSS, SIZE(DU,4), SIZE(SS,4), self%km, self%klid, &
                             self%cdt, MAPL_GRAV, fMassHNO3, fMassNO3, NO3an1, NO3an2, & 
                             NO3an3, HNO3CONC, HNO3SMASS,  HNO3CMASS, __RC__)
```

pcolarco:
Well, this will solve that immediate problem. However it doesn't really address the hard-wiring. There is a presumption in the code that maps the heterogeneous rates calculated for dust and sea salt per bin to production terms to the 3 nitrate size bins. That requires a more thoughtful solution to properly map (arbitrary) dust and sea salt bins appropriately. But I suppose what you've got now doesn't crash.

amdasilva:
[...]

PR #175, opened 2022-08-22
Removed nbinsDU and nbinsSS arguments from subroutine NIheterogenousChem

weiyuan-jiang:
This PR addresses issue #171


amdasilva (approved):
Although code no longer gives NaNs, it still may not work correctly. It has no effect for current GOCART-2G configuration at GMAO, so going with it for now. A better fix needs to be implemented.

amdasilva:
Will need better fix later on.

Issue #70, opened 2021-08-05
Use DSOs?


tclune:
During the development of GOCART2G,  we updated the rest of the GCM to use DSOs and thereby eliminated compile time dependencies between components.    GOCART2G has not been updated in that manner.

Should this be done _before_ v2.0.0 or after?

amdasilva:
Be my guest to implement DSOs, wthatever that is... I am assuming this is a cmake level update.

amdasilva:
@tclune Is this something to be done now?

tclune:
@amdasilva I think we _can_ do it with relative ease at any point.  Relatively simple CMake surgery.

@atrayano did recently do something like this with MIT ocean, and got different answers which gave us pause; so we'd best double check that there are no surprises with GOCART.   

I can add it to the backlog for @weiyuan-jiang

weiyuan-jiang:
There seems already no dependency of gorcar2g on its children
https://github.com/GEOS-ESM/GOCART/blob/4c941791d3d1fbea65dd5d2be330ada8c4951036/ESMF/GOCART2G_GridComp/CMakeLists.txt#L24-L29

weiyuan-jiang:
@tclune , what is the specific issue here?

tclune:
The use statements here should go away: 

https://github.com/GEOS-ESM/GOCART/blob/4c941791d3d1fbea65dd5d2be330ada8c4951036/ESMF/GOCART2G_GridComp/GOCART2G_GridCompMod.F90#L20-L24

Need to use add child via DSO interface.

PR #182, opened 2022-09-08
Feature/wjiang/shared lib

weiyuan-jiang:
This PR is addressing issue #70 . The components of GOCART2G are built as DSO.   Do we need to build  GOCART2G as DSO too ? @tclune

### DIFF
--- a/ESMF/GOCART2G_GridComp/GOCART2G_GridCompMod.F90
+++ b/ESMF/GOCART2G_GridComp/GOCART2G_GridCompMod.F90
@@ -1045,7 +1045,7 @@ contains
 
     ! locals
     integer                                                    :: i
-    character(:), allocatable :: dso_suffix
+    character(:), allocatable :: dso_libname
 
     __Iam__('GOCART2G::createInstances_')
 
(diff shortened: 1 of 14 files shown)


ESMF/GOCART2G_GridComp/GOCART2G_GridCompMod.F90
```diff
@@ -1062,22 +1055,30 @@ subroutine createInstances_ (self, GC, rc)
 !   Active instances must be created first! This ordering is necessary for
 !   filing the AERO states that are passed to radiation.
 !   This is achieved by arranging the names of the active instances first.
+    dso_libname = adjust_dso_name('libDU2G_GridComp')
```
mathomp4, line 1058: @weiyuan-jiang You probably don't need to do this. MAPL_AddChildFromDSO should adjust it for you automatically. So you could just pass in, say, `libDU2G_GridComp.so`.  But maybe this is better? Hmm.

That said, note that the CI is failing because I'm not sure we expose `adjust_dso_name` publicly in MAPL. Might be a question for @tclune if it should be or not.

weiyuan-jiang: Ahh, you are right. It is adjusted inside the program. I will clean this up

tclune: This exercise is in part to expose issues at various stages of build/run.    Relocatability is the big concern that a poorly understand.

amdasilva:
We need to revisit this later. We need a better plan to split the process library.  I will close this without merging it.

PR #187, opened 2022-09-16
Fixes #184. Remove unneeded GFE find_package calls

### LINKED ISSUE #184: Unnecessary yaFyaml dependency in GOCART build system
yaFyaml is not needed to build GOCART. However, the following dependency is included in the build system:
https://github.com/GEOS-ESM/GOCART/blob/e59ddd490f3bc413465317a89e5684cee68de8e8/CMakeLists.txt#L64

mathomp4:
Closes #184 

This removes some unneeded `find_package` calls for GFE libraries. These are needed by MAPL not GOCART and so are redundant here (and as @rmontuoro pointed out, bad).

This also adds a new CI test to build GOCART like UFS.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,6 +24,21 @@ workflows:
           mepodevelop: true
           develop_repos: GMAO_Shared
           persist_workspace: false # Needs to be true to run fv3/gcm experiment, costs extra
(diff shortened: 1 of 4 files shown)


amdasilva (approved):
Getting this to develop makes sense.

PR #188, opened 2022-09-20
Update CI BCs version to fix GEOSgcm run

mathomp4:
Due to recent changes in the GEOSgcm_App, we need to advance the BCs version in the CI so that the scripting for making an experiment works.


amdasilva (approved):
Looks good.

Issue #183, opened 2022-09-14
Add various aerosol exports 


vbuchard:
- add exports for total extinction and scattering coefficient profiles needed for JEDI exp and chemistry group
- add total (molecular + aerosols) single scattering attenuated backscatter coefficient for PBL modeling group

PR #189, opened 2022-09-23
Feature/vbuchard/extinction export

vbuchard:
- added exports for total extinction and scattering coefficient profiles at model RH and at fixed RH=20% and 80%
- Partially resolved issue #183


amdasilva (dismissed):
Looks reasonable.

PR #198, opened 2022-11-30
Feature/vbuchard/total aback toa

vbuchard:
Adds:
- Aerosol single scattering backscatter coefficient for each instances and total at wavelengths_profile
- Total (molecular + aerosols) attenuated backscatter coefficient from TOA and sfc at 532nm.
This PR closes issue #183. The total attenuated backscatter coefficient calculation will be revisit later to include other wavelengths than 532nm.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Extinction exports with RH=20% and RH=80%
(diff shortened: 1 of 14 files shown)


amdasilva (approved):
This feature appears to work with a few caveats. It requires that dependent exports have been activated, for now my means of an entry in history. This is undesirable because of the unnecessary extra I/O. 

Solution: a request have been made to the SI Team to extend MAPL so that export dependencies are specified during AddExportSpec.  As of this writing, work was in progress by Atanas.

PR #205, opened 2023-01-09
Fixes #202. Add * to Long Name for CA Gridded Component

### LINKED ISSUE #202: Update State Specs for *-expansion in Long Names
@darianboggs recently added the ability in MAPL to expand * in long name with the ACG:

https://github.com/GEOS-ESM/MAPL/pull/1881

This feature will be in the next version of MAPL (2.34.0) which will be released "soon". As such, this is an issue reminder to update the state specs in files where star-expansion is done. At the moment I think it's only in CA:

https://github.com/GEOS-ESM/GOCART/blob/984fc494abfcd95ff3c2ced49dc0ab11176b8510/ESMF/GOCART2G_GridComp/CA2G_GridComp/CA2G_StateSpecs.rc#LL75-L78

but others might use it in branches in development? The above lines should become:
```
 NAME         | UNITS        | DIMS  | VLOC  | UNGRIDDED                      | LONG NAME
#----------------------------------------------------------------------------------------
 *MASS        | kg kg-1      | xyz   | C     |                                | * Carbonaceous Aerosol Mass Mixing Ratio
 *CONC        | kg m-3       | xyz   | C     |                                | * Carbonaceous Aerosol Mass Concentration
```
as an example. Note you want to make sure you have the space so this:
```
* Carbonaceous ...
```
not:
```
*Carbonaceous ...
```
as the Python will do:
```
CA.bcCarbonaceous ...
```
for the latter.

mathomp4:
Closes #202 

With MAPL 2.34 now in GEOSgcm `main`, I feel confident about making this PR. To wit, this PR uses the updated *-expansion in MAPL's ACG to expand out the Long Name in the State Specs much like it does in the short name. Note that this is a bit different than might be expected[^compname].

Note I also changed the Internal state *-variables here:
```
 *phobic | kg kg-1 | xyz  | C    | MAPL_RestartOptional | T          | DYNAMICS:TURBULENCE:MOIST | Carbonaceous Aerosol Mixing Ratio
 *philic | kg kg-1 | xyz  | C    | MAPL_RestartOptional | T          | DYNAMICS:TURBULENCE:MOIST | Carbonaceous Aerosol Mixing Ratio
```
to:
```
*phobic | kg kg-1 | xyz  | C    | MAPL_RestartOptional | T          | DYNAMICS:TURBULENCE:MOIST | * Hydrophobic Aerosol Mixing Ratio
*philic | kg kg-1 | xyz  | C    | MAPL_RestartOptional | T          | DYNAMICS:TURBULENCE:MOIST | * Hydrophilic Aerosol Mixing Ratio
```

because these are also added to Export (the `T` column). Thus, history output will have the right name. But because these are *Internal* state variables and changing could trigger a "non-zero-diff" in checkpoints. Now, It would only be the `long_name` attribute and not the actual data, but I wanted to be conservative.

I also took the time to update the `components.yaml` to match GEOSgcm `main` so it uses MAPL 2.34 as well.

[^compname]: In #202 (or rather the chain leading to it, see https://github.com/GEOS-ESM/MAPL/issues/1877), @amdasilva said:  
    > image
    > 
    > For the short name the “*” is replaced with the name of the component, thus BCEXTTAU, BRCEXTTAU, etc. A similar device was supposed to have been implemented for the long name but apparently it has not. In this case, we should have something like this:
    >  
    > Which would expand to “BC Extinction AOT”, “BRC Extinction AOT”, etc… Expanding BC to “Black Carbon” cannot be easily done with the ACG without introducing cumbersome syntax. Having ncks perform attribute renaming at the script level may be easier to implement.
    
    It turns out technically we cannot do this. When you use MAPL 2.34 with this PR, you get out:
    ```
    BCEM001:long_name = "CA.bc Aerosol Emission (Bin 001)" ;
    ```
    for, say, `BCEM001`. Note this does *NOT* say `"BC Aerosol Emission (Bin 001)"`. The reason is that this variable internally is `CA.bcEM001` and not `BCEM001` because the component is `CA.bc`; it's called `BCEM001` only because of aliasing in `HISTORY.rc`:
    ```
    'CA.bcEM'        , 'CA.bc'     , 'BCEM001;BCEM002' ,
    ```
    But, as far as the ACG is concerned, this is `CA.bc` and it has no way to know that someone renamed it in `HISTORY.rc`.
    
    I believe GEOS will need a postprocessing step with something like `ncatted` to convert `CA.bc` to either `BC` or `Black Carbon`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `*` to CA State specs file to allow for ACG to substitute in the instance name
+
 ### Fixed
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Moved to use GitHub Action for label enforcement
 - Updated CircleCI image to use Baselibs 7.7.0
 - Update `components.yaml` to reflect GEOSgcm
+- CA restarts will have a change in longname for `philic` and `phobic` variables due to addition of `*` in the CA State specs file
+  for the Internal state variables
 
 ## [2.1.2] - 2022-10-07
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Added `*` to CA State specs file to allow for ACG to substitute in the instance name
+- Added `*` to CA State specs file to allow for ACG to substitute in the long name
 
 ### Fixed
 
--- a/ESMF/GOCART2G_GridComp/CA2G_GridComp/CA2G_StateSpecs.rc
+++ b/ESMF/GOCART2G_GridComp/CA2G_GridComp/CA2G_StateSpecs.rc
@@ -113,8 +113,8 @@ category: INTERNAL
 #--------------------------------------------------------------------------------------------------------------
   NAME   | UNITS   | DIMS | VLOC | RESTART              | ADD2EXPORT | FRIENDLYTO                | LONG NAME
 #--------------------------------------------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


amdasilva (dismissed):
Seems reasonable.

amdasilva (commented):
Good to go.

amdasilva (approved):
Approved.

PR #214, opened 2023-03-10
Merging refactored physics code into develop

sdrabenh:
NOTE: This branch is *LOCKED*. This is because it was based on v2.1.2 not `develop` (which is like v2.1.3) so if we "Update branch" we would bring in additional untested non-zero-diff changes.


amdasilva:
Scott will work from latest develop.

PR #224, opened 2023-04-05
Fix 3 hourly frequency update in go-cart to allow start/stop regression in less than 3 hour chances

### LINKED ISSUE #146: Hard coded 3 hourly frequency in gocart2g
@amdasilva @pcolarco @adarmenov  @weiyuan-jiang @tclune @mathomp4 

There is this issue reported in UFS that appears to involve gocart2g:
https://github.com/ufs-community/ufs-weather-model/issues/1190

now Matt noticed that a 4 hour vs 2 + 2 start/stop regression was failing. We then remembered there is a hard coded 3 hourly alarm in gocart2g! In these spots:
https://github.com/GEOS-ESM/GOCART/blob/v2.0.6/ESMF/GOCART2G_GridComp/NI2G_GridComp/NI2G_GridCompMod.F90#L393
https://github.com/GEOS-ESM/GOCART/blob/v2.0.6/ESMF/GOCART2G_GridComp/SU2G_GridComp/SU2G_GridCompMod.F90#L480

If I change this to 2 hours, the 4 hour vs 2 + 2 regression passes. So it does seem possible this could be the problem for UFS. At the very least it seems like either whatever the point of this alarm is it either needs to be divisible by the job segment or perhaps we are not saving something in the internal state that needs to be checkpointed.

bena-nasa:
fixes #146

This fixes a long standing issue that one can not start and stop the model in anything less than 3 hour increments to test start/stop regression because of GOCART

I will not rehash why this occurring other than to say this is unacceptable. If you look in the issue you will see all the details. The one thing I had to do was remove ESFM alarms, the whole problem was that an alarm was made relative to the model start time which just can not be. But the problem is if the ESMF Alarm is created correctly, say at 0z of the day so that no matter when you start you get the same behavior, it does not work right as there are fundamental bugs in ESMF alarms. ESMF is aware of this and will fix it some day, but until then we must avoid using them. In it's place I made a simple function that says is my time some frequency from 0z during the day.

This is zero-diff in my test and allows one to do start/stop regression at less than 3 hour increments which is very, very useful for debugging when you need it and multiple people have wasted time doing regression tests and were not aware of this "bug". Plus this what was coded was fundamentally not right.

Note I made the same function in two routines as I did not see any common place in GOCART2G for helper procedures such as this to go, although if there is, or we should invent one, I can do that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - To do: remove hooks to old (legacy) GOCART.data instances in CHEM and setup scripts
 - Fixed rc file in legacy O3 component.
 - Fixed issue #223 where Global dimension was being used for allocating a local array	
-- 
+-    
 ### Added
 
 ### Changed
--- a/ESMF/GOCART2G_GridComp/NI2G_GridComp/NI2G_GridCompMod.F90
+++ b/ESMF/GOCART2G_GridComp/NI2G_GridComp/NI2G_GridCompMod.F90
@@ -767,7 +767,6 @@ contains
     if (alarm_is_ringing) then
        xhno3 = NITRATE_HNO3
     end if
-    if (mapl_am_i_root()) write(*,*)"bmaa ni alarm ring ",alarm_is_ringing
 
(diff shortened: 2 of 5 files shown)


amdasilva (dismissed):
Looks good.

vbuchard (approved):

Issue #223, opened 2023-04-05
Global dimension for a local array


aoloso:
In GOCART2G children, globalCellCountPerDim is used instead of localCellCountPerDim to dimension local arrays, leading to excessive memory wastage. Changes were made to use localCellCountPerDim.

PR #225, opened 2023-04-05
Issue #223 Global dimension for a local array

aoloso:
In GOCART2G children, globalCellCountPerDim is used instead of localCellCountPerDim to dimension local arrays, leading to excessive memory wastage. Changes were made to use localCellCountPerDim.


amdasilva (dismissed):
This is a very important patch to prevent unnecessary allocation of global arrays.

vbuchard (approved):

PR #234, opened 2023-05-15
Fix for totscacoefrh80 and OC

mathomp4:
It looks like a bug might have come into GOCART a bit ago when backscatter calculations were added. If you look at the changes here, the `totscacoefrh80` with `ocscacoefrh80` code was transformed into a *second* `totscacoefrh20` with `ocscacoefrh20` calculation.

I believe this is a bug, as the 20/80 duo exist for other species.

cc: @vbuchard


vbuchard (approved):
Good catch! Thank you Matt

amdasilva (approved):
Changelog conflict resolution

PR #238, opened 2023-05-18
Added support for neural network prediction of size distribution

dbarahon:


amdasilva:
Donifan: Thank you for the contribution of MAMNET, this is something that could be very useful. Please add a description, and make sure all CI tests passes.   We are having a bit difficulty understand the code. Would you mind meeting with us and give a code walk thru? Since we will be maintaining this code, it is important that we fully understand what you did.

PR #240, opened 2023-05-30
Fixes #239. Use global dims for Chem_UtilResVal

### LINKED ISSUE #239: Release v2.2.0 sees an increase in dust and sea salt emissions
- Just a notification that GOCART Release v2.2.0 sees an increase in dust and sea salt emissions (~ factor 2) compared to previous version that is currently under investigation.

mathomp4:
Closes #239 

This is a hotfix onto `main` to try and fix #239 by passing global dimensions into Chem_UtilResVal (as it expects global dims). 

Labeling as non-zero-diff as it would change answers though probably a bugfix "good" diff.

I'll keep draft until @vbuchard can test. 

Also note to @sdrabenh, this bug is probably in sdr_v2.1.2.6 as well, so we'd need a new tag there for v11

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fix dims passed to `Chem_UtilResVal` as they must be global (see #239)
+
 ### Added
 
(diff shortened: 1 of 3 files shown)


amdasilva:
I have a separate PR addressing the same issue.

PR #243, opened 2023-06-08
Update components to match GEOSgcm v11.1.0, update CI

mathomp4:
This PR updates the "in-repo" components to match GEOSgcm v11.1.0. This is mainly used for build testing of GOCART without the full model.

This PR also updates the CI to point to Baselibs 7.13.0 which is the version used by GEOSgcm v11.1.0. 

Note that for GEOSgcm, Baselibs 7.13.0 updates to ESMF v8.5.0b22 which has a non-zero-diff bugfix to grid generation (precision fix).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 # Anchors to prevent forgetting to update a version
-baselibs_version: &baselibs_version v7.7.0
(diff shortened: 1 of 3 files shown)


amdasilva (approved):
It looks good, it does not affect source and has no overall impact on science.

PR #246, opened 2023-07-24
Enables Climatological Anthropogenic Emissions after 2019

acollow:
This PR modifies the "_ExtData.yaml" files for CA, NI, and SU to persist 2019 anthropogenic emissions as a climatology. Note this version of the PR does NOT include HEMCO diurnal and DoW scaling. During leap years, this will not be zero diff to the previous methodology of copying emissions files with future years in the file name because ExtData2G will update the index of the file one day later. Otherwise, the functionality is the same.


jrjoshi1 (approved):
Arlindo, Virginie, and I went through the code change and the changes look reasonable. We consider it a 0-diff, although small difference may exist regarding leap years. However, concern still remains that persisting 2019 is not a good idea due to pandemic.

vbuchard (approved):

PR #253, opened 2023-09-15
Fix/fengsha moist

bbakernoaa:
Found an issue with the gravimetric soil moisture units used within the FECAN soil moisture correction.  This fixes that as well as adds a few knobs to be able to tune the soil moisture and drylimit within the dust scheme.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [2.2.2] - 2023-09-19
+## [Unreleased]
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,8 +18,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Correct soil moisture parameterization in FENGSHA
-- Add `soil_moisture_factor` to the DU2G_GridComp.rc and DU2G_GridCompMod.F90 files for FENGSHA
-- Add `soil_drylimit_factor` to the DU2G_GridComp.rc and DU2G_GridCompMod.F90 files for FENGSHA
+- Add `soil_moisture_factor` to the DU2G_instance_DU.rc (same name used in the K14 scheme) and DU2G_GridCompMod.F90 files for FENGSHA
+- Add `soil_drylimit_factor` to the DU2G_instance_DU.rc and DU2G_GridCompMod.F90 files for FENGSHA
 
 ## [v2.2.1] - 2023-05-30
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,13 +5,21 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [2.2.2] - 2023-09-19
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


bbakernoaa:
Anything I need to do for this?

mathomp4:
@bbakernoaa A couple things:

1. Can you add an entry to the CHANGELOG.md file in the root directory explaining your changes.
2. We require a label (either 0-diff or non-zero-diff) to let the @GEOS-ESM/aerosol-team team know if this would affect results. Now, technically I suppose it is non-zero-diff if you run Fengsha but then we (GMAO) don't run that I think, so...I guess you can label it zero-diff? (Since, in a way, it is to us.)

bbakernoaa:
@mathomp4 I'll try to add something to the changelog and the label this morning.  Thanks

CHANGELOG.md
```diff
@@ -5,14 +5,22 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [2.2.2] - 2023-09-19
```
mathomp4, line 8:
```suggestion
## [Unreleased]
```

mathomp4: @bbakernoaa Let's keep it as unreleased. We don't know when the PR will go in and what the release number will be or the like.

mathomp4:
> @mathomp4 I added a changelog but if I'm suppose to add a label here in the PR I don't have permission to do so.

Ah. Interesting. I didn't know you needed write access to make a label! Well, I've done it for you. I'll keep it as zero-diff since we don't run Fengsha.

But, I'll be careful and do a test to make sure!

mathomp4:
Yep. This seems to be zero-diff at least how I know how to run GOCART (which I guess isn't Fengsha!)

bbakernoaa:
Good this should only effect FENGSHA and not the Ginoux or K14 schemes

tclune:
@amdasilva  @vbuchard Our NOAA colleagues would like to know when they can expect this to be merged.

mathomp4:
@bbakernoaa We are confused by:
```
- Add `soil_moisture_factor` to the DU2G_GridComp.rc and DU2G_GridCompMod.F90 files for FENGSHA
- Add `soil_drylimit_factor` to the DU2G_GridComp.rc and DU2G_GridCompMod.F90 files for FENGSHA
```

First, do you mean `DU2G_instance_DU.rc`? Second, on your branch, I see no updates to the FENGSHA bits at the bottom:

https://github.com/bbakernoaa/GOCART/blob/fix/fengsha_moist/ESMF/GOCART2G_GridComp/DU2G_GridComp/DU2G_instance_DU.rc

there is only:
```
# FENGSHA settings
alpha: 0.3
gamma: 1.3
vertical_to_horizontal_flux_ratio_limit: 2.e-04
```

bbakernoaa:
@mathomp4 Yes sorry.  It will need to be in the `DU2G_instance_DU.rc` file.  I used the same names as the k14 so I didn't want to duplicate it in the `rc` file. They do the same things as in the k14 scheme. 

Could you please advise on how we should do that?

mathomp4:
> @mathomp4 Yes sorry. It will need to be in the `DU2G_instance_DU.rc` file. I used the same names as the k14 so I didn't want to duplicate it in the `rc` file. They do the same things as in the k14 scheme.
> 
> Could you please advise on how we should do that?

Ohhh. Okay. I see. I didn't notice they were above and not commented out! Well, I suppose this works then.

vbuchard:
> @mathomp4 I just added the soil_drylimit_factor to the `.rc` file as this one wasn't duplicated

Can you please update the changelog accordingly (with the right filename) before I can approve/merge the PR. With all these updates, it looks ok to me. Thanks @mathomp4 for doing the testing.

bbakernoaa:
@vbuchard I modified it.  Please let me know if that works.

vbuchard (approved):
It looks good to me. Thanks

PR #261, opened 2024-01-08
prc: updates to include separate explosive and degassing volcanic emissions for SO2

pcolarco:
Modification to SU2G_instance_SU.rc to now provide two input files for volcanic SO2 emissions, one for explosive and one for degassing. Moved existing volcanic emissions to one or the other. Updated Fortran code in SU and Process_Library to accommodate this change. Verified zero-diff result for tracer and restart against v11.3.3 release of GEOSgcm. This is not presently zero-diff in volcanic emission diagnostics, as presently they are assigned to whichever of the channels I put the emissions to. This will be corrected when we update the emissions next go round.


jrjoshi1:
@pcolarco Please update PR to develop instead to main, and resolve any conflict that may arise.

pcolarco:
Joshi, can you show me how to do this? I do this so infrequently (this was 8 months ago...) and clearly I don't remember to do it the right way.

jrjoshi1:
> > Joshi, can you show me how to do this? I do this so infrequently (this was 8 months ago...) and clearly I don't remember to do it the right way.

@pcolarco I am not sure but will keep you updated.

mathomp4:
The first thing to do is go to the top of the PR where there is an "Edit" button. When you click that, under the title bar, you'll see a new dropdown for "base". Change that to `develop` and click Save and see what happens. You'll probably get an "Update branch" button at the bottom to click.

Once you press that, it might merge in clean, or you might have to fix up some conflicts.

mathomp4:
I can work with either of you on this if you like :)

PR #264, opened 2024-02-02
Feature/pcolarco/#263/su2g_gridcompmod logic

pcolarco:
These are several enhancements captured here:

Sulfate surface area density calculation in SU_Compute_Diags was incorrectly being passed the effective radius used for settling along with the sigma width of the number distribution. Properly it should be passed the number median radius, also present in the RC file. Added a hook to read that field from the RC file ("particle_radius_number"), store in SU grid comp, and pass to SU_Compute_Diags. This change is zero-diff to the SU internal state. It changes value of export SO4AREA.
Add some protective logic around reading daily volcanic emission files. If filename does not exist model will write message to standard output and reset volcanic emissions to zero. This is zero difference result unless in a place where the files were not present.
Changed DMS concentration data holder from ExtData provided (SU_DMSO) to local copy (dmso_conc). This is relevant since if we run source tagged instances where we don't want DMS emissions we would zero out dmso_conc and that is what should be passed to DMSemission subroutine. This is zero diff except in that case.


jrjoshi1:
@pcolarco we merged #261 into develop. Now we are having conflict in SU2G_GridCompMod.F90. Could you please take a look into it and see if you can resolve the conflict easily between develop and your branch. Thanks.

pcolarco:
I believe this is merged properly.

mathomp4:
Yay! All tests passed!

jrjoshi1 (approved):
looks okay.

CHANGELOG.md
```diff
@@ -60,6 +60,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `soil_drylimit_factor` to the DU2G_instance_DU.rc and DU2G_GridCompMod.F90 files for FENGSHA
 - Moved process library macros to header file.
 
+## [#263] - 2024-02-02
+
+### Changed
+-Sulfate surface area density calculation in SU_Compute_Diags was incorrectly being passed the effective radius used for settling along with the sigma width of the number distribution. Properly it should be passed the number median radius, also present in the RC file. Added a hook to read that field from the RC file ("particle_radius_number"), store in SU grid comp, and pass to SU_Compute_Diags. This change is zero-diff to the SU internal state. It changes value of export SO4AREA.
+
+-Add some protective logic around reading daily volcanic emission files. If filename does not exist model will write message to standard output and reset volcanic emissions to zero. This is zero difference result unless in a place where the files were not present.
```
jrjoshi1, line 68: Does this part still holds relevance because the change in the code does not seem to indicate a test for the file's existence?

vbuchard (approved):
Note: During merging, this enhancement is no longer part of this PR (will need testing)  "Add some protective logic around reading daily volcanic emission files. If filename does not exist model will write message to standard output and reset volcanic emissions to zero. This is zero difference result unless in a place where the files were not present."

jrjoshi1 (approved):

mathomp4 (approved):

PR #273, opened 2024-04-17
Fix CMake circular dependencies for external use

mathomp4:
As found by @climbfuji, if you build GEOSgcm, you get three sets of CMake files for doing `find_package`: MAPL, GEOSgcm, and GOCART. This then causes circular dependencies as GOCART requires things "provided" by GEOSgcm and GEOSgcm requires GOCART.

So this PR tries to "detect" if GOCART is being built as a standalone project or as part of GEOS.

ETA: In tests trying to build GOCART with MAPL-as-library-from-spack, it was found that you need to protect against the chance MAPL is found as a library. Then you can't *add* MAPL as a new subdirectory as targets will be duplicated.


mathomp4:
@climbfuji said this worked for him. So I'll undraft.

rtodling:
I hope this can make up so 11.5.3 version of the model - very important for GEOS-JEDI development.

climbfuji (approved):
Thanks for thes updates @mathomp4 . This works fine for us (JCSDA)!

climbfuji (approved):
I don't understand these changes very well, but if it works it works!

vbuchard (approved):

PR #282, opened 2024-07-19
Addresses the incorrect seasonal cycle in the CEDS emissions for the …

acollow:
…current x run testing


mathomp4 (approved):

mathomp4:
@acollow We've tagged `sdr_v2.3.0` with this update

PR #287, opened 2024-08-19
Fix issue between UFS build and standalone CMake

mathomp4:
As found by @DusanJovic-NOAA in https://github.com/ufs-community/ufs-weather-model/issues/2399 my previous CMake PR (#273) worked well for us, but broke UFS. Mea culpa.

So, we add CMake to say "if `UFS_GOCART` is true, so is `GOCART_STANDALONE`"


vbuchard (approved):
Thanks Matt.

PR #294, opened 2024-10-10
SU_Volcanic: file path and protection for missing file

jrjoshi1:
Degassing file path is corrected. Additionally, a safeguard has been implemented to halt model execution if the specified volcanic file does not exist for the requested date.

### DIFF
--- a/ESMF/GOCART2G_GridComp/SU2G_GridComp/AMIP/SU2G_instance_SU.rc
+++ b/ESMF/GOCART2G_GridComp/SU2G_GridComp/AMIP/SU2G_instance_SU.rc
@@ -9,7 +9,7 @@ nbins: 4
 
 # Volcanic pointwise sources
 volcano_srcfilen_explosive: ExtData/chemistry/CARN/v202401/explosive/so2_explosive_volcanic_emissions_CARN_v202401.%y4%m2%d2.rc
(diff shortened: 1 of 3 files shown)


mathomp4:
@jrjoshi1 I don't think we can do this. Your change does stop the model:
```
 Error: file not found
 ExtData/chemistry/CARN/v202401/explosive/so2_explosive_volcanic_emissions_CARN_
 v202401.20000414.rc
```
but we don't want that. We just want to model to handle a missing date gracefully. For example, your code means the model cannot run save on 19 days in the year 2000.

I think we just don't need the `else` bit. Just if the file exists, read it. If not, it should just avoid the volcano codes.

jrjoshi1:
@mathomp4 I anticipated that concern! Without the 'Else and stop,' the model might continue running, leading users to believe that all volcanic files were utilized for the simulation period, while some files, like those for 2000, may not be included. If we're confident that no files are/(will be) missing for any year except 2000, we could consider removing the 'Else/stop' part; however, users should still be aware of the issue. Just my thoughts! 
Don't know what @pcolarco and others would think.

jrjoshi1:
@mathomp4 I see your point now, and I think we should remove the 'else/stop' part. Unless there are any other opinions, I’ll go ahead and make that change tomorrow. Thanks.

mathomp4:
This seems to work for me. I'll approve since it's it is a pretty important bugfix.

mathomp4 (approved):

Issue #301, opened 2025-02-11
Update


amdasilva:
# Background

When implementing the AOP (Aerosol Optical Properties) module in pyobs it became evident that existing ordering of the array dimensions in the netcdf Mie LUT tables was somewhat inefficient. This prompted the adoption of a revision of these tables.

The versions of the LUT does not refer to the revision of the values in each table, but rather to the structure of the netCDF files containing these tables

v0.0.0: original file structure, e.g.

	qext(radius, rh, lambda) 
	pmom(nPol, nMom, radius, rh, lambda) 
				
v1.0.0: dimension ordering has been updated for cache optimization (use case: radiation and RT calculations). 
            The following dimensions have been renamed: lambda --> wavelength, radius --> bin.
			
	qext(bin,wavelength, rh)
	pmom(bin,wavelength, rh, p, m)

This utility can be used to convert v0.0.0 files to v1.0.0:

https://github.com/GEOS-ESM/GMAOpyobs/blob/develop/src/utils/aop_0to1

# New Tables on Discover

Optical tables v1.0.0 can be found on discover at:

   /discover/nobackup/projects/gmao/share/dasilva/fvInput/ExtData/chemistry/AerosolOptics/v1.0.0

# Update GOCART to work with MieTables v1.0.0

The Fortran code in `Process_Library/GOCART2G_MieMod.F90` needs to be updated to work with the dimension ordering of tables v1.0.0.

weiyuan-jiang:
I cannot find the variable qext in the develop branch. @amdasilva

weiyuan-jiang:
@amdasilva ,  Is it just pmom ? I think all the variables should change.

weiyuan-jiang:
In this data structure block, all the variables' dimension reflecting the order in the file.  Should we change the pmom too ?[pmom
](https://github.com/GEOS-ESM/GOCART/blob/7cf6f6007cccd2f814e72b065d4aa1734c6fbea6/Process_Library/GOCART2G_MieMod.F90#L65)

I suggest pmom ( m, p, r, c, b) @amdasilva

amdasilva:
@weiyuan-jiang Just now I  saw your questions questions. I typically filter out all GitHub email. Answering your questions:

* I cannot find the variable qext in the develop branch
  - Answer: this variable is in the file, it could have been an omission
* Is it just pmom ? I think all the variables should change.
   - Answer: Yes, ALL variables must be changed to conform to the changes in the dimension ordering
[...]

PR #302, opened 2025-02-14
reordering dimension

weiyuan-jiang:
This PR addressed issue #301

### DIFF
--- a/ESMF/GOCART2G_GridComp/CA2G_GridComp/CA2G_instance_CA.bc.rc
+++ b/ESMF/GOCART2G_GridComp/CA2G_GridComp/CA2G_instance_CA.bc.rc
@@ -2,7 +2,7 @@
 # Resource file for Black Carbon parameters. 
 #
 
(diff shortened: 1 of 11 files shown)


amdasilva:
Have you tested it to very zero diff?

amdasilva (changes requested):
Let's talk, I don't think this is correct.   The dimensioning of the arrays in the Fortran code should be changed to match the new order in the NetCDF file (of course, the data layout is what should should be the same, the order of dimensions is reversed because of the different ways Fortran and NetCDF/C index arrays.) It appears that you are reordering the array read from the file, and keeping the old Fortran ordering.

weiyuan-jiang:
This PR is zero diff with this branch  feature/wjiang/dev_for_compare , which is slightly different from the develop branch.  In develop branch, it uses some files from  /discover/nobackup/pcolarco/fvInput/AeroCom/x/  which is not reordered

patricia-nasa:
These changes were pulled in with another PR

PR #319, opened 2025-05-05
set hsurf according to the index range

weiyuan-jiang:
Try to address the comments here:

https://github.com/GEOS-ESM/GOCART/blob/c393e57d4472c06ec61b770f8f46a152fd3280ff/Process_Library/GOCART2G_Process.F90#L1182


patricia-nasa:
NOAA has their own chem settling routine.  These changes not needed anymore.

PR #336, opened 2025-06-04
Feature/pcolarco/chem coupling

pcolarco:
This adds hooks in SU2G_GridComp to on-way couple GMI oxidants (OH, NO3, H2O2) to GOCART sulfur chemistry. Control is through instance RC file; off by default
Add "disable_emissions" feature to SU2G_instance file to turn off sulfur emissions except what is specified in instance RC; useful for volcanic only instance
Add calculations of surface area density and effective radius to GOCART sulfate and carbonaceous species, and add to respective aerosol states for use in chemistry
Made a change in MieQuery.H to *not* convert units of effective radius in place to [um]. Units are now read in native units on optics file [m] and made corresponding change in Chem_Settling call
Made updates to latest version of IDL legacy optics tables for surface area and effective radius calculations
All changes are zero diff except the optics tables


pcolarco:
Added CHANGELOG update

PR #338, opened 2025-06-10
Added logic to ensure the alarm accounts for skipping heartbeat to develop

mathomp4:
This is the equivalent of #337 but to `develop` instead of `feature/sdrabenh/gcm_v12`

Per that PR via @atrayano:

> This fixes the behavior of the daily_alarm subroutine (used to control replenishment of aerosols) for NI and SU components when GOCART2G does not at every heartbeat of the clock

I'm not sure if zero-diff or not...so I'll label both.


vbuchard:
Thanks @mathomp4 for helping merging this PR into develop! Do you know why 2 tests are failing?

mathomp4:
> Thanks @mathomp4 for helping merging this PR into develop! Do you know why 2 tests are failing?

Ugh. Yes. Let me push a fix onto `develop`.

PR #342, opened 2025-06-26
Bring in nco changes for DEBUG intel flags

BrianCurtis-NOAA:
Bring in NCO DEBUG flag change for ops

The following flags will be added for DEBUG builds: -check all -check noarg_temp_created -ftrapuv

Two are required by NCO: -check all -ftrapuv
One will drastically reduce build output file size: -check noarg_temp_created

I've tested the build on the UFSWM. I will run the full UFSWM RT suite soon to ensure the build change does not impact results. I do not anticipate any changes.


BrianCurtis-NOAA:
I do. I've had a few front-burner items that have pushed this to the back. As soon as I can coordinate this effort across all UFSWM components, I'll  move it off draft-mode. Apologies to keep this waiting.

mathomp4:
> I do. I've had a few front-burner items that have pushed this to the back. As soon as I can coordinate this effort across all UFSWM components, I'll move it off draft-mode. Apologies to keep this waiting.

No problem. My fear was this was meant to be not-draft long ago and was forgotten, etc. It doesn't affect us so we should be able to take it at anytime.

mathomp4:
> Will the latest develop changes cause baseline differences? I'm seeing some, and wanted to verify that any tests using GOCART will have results changing.

@BrianCurtis-NOAA GOCART `develop` is non-zero-diff to GOCART v2.4 in our testing, yes, though I can only say from the GEOSgcm perspective. From the diffs:

https://github.com/GEOS-ESM/GOCART/compare/main...develop?w=1

You can see that a pressure lid is now used in many places (say 40 mbar in black carbon). The Ginoux scheme and other scalings also had updates for L181 which we are moving to. We also updated some optics files. Finally, `SU_Compute_Diags` seems to have had a unit fix/change.

I'll mention @vbuchard @pcolarco and @jrjoshi1 for more edification.

BrianCurtis-NOAA:
@mathomp4 We'll need to look at restart reproducibility with the latest GOCART in UFSWM. We had to ditch it. My changes will not change results, so it's still OK to be in GEOS-ESM/develop.

PR #344, opened 2025-06-30
Bring over latest Atanas alarm fix to develop

mathomp4:
This is the equivalent of #343 but to `develop`. 

This has fixes by @atrayano for the alarms in GOCART in NI and SU when running replay. 

It is non-zero-diff in my testing.

### DIFF
--- a/ESMF/GOCART2G_GridComp/GOCART2G_GridCompMod.F90
+++ b/ESMF/GOCART2G_GridComp/GOCART2G_GridCompMod.F90
@@ -187,6 +187,8 @@ contains
 !   -----------------------------------------------------------------
     call createInstances_(self, GC, __RC__)
 
+    call alarmResourcesToChildren(self, GC, _RC)
+
 !   Define EXPORT states
 
 !   This state is needed by radiation and moist. It contains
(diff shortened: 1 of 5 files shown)


acollow:
@mathomp4, can you clarify on what was non-zero diff about this? I can certainly test this but will need some help creating a branch analogous to GOCART 2.5-rc1 with this change implemented on top, and instructions on how to build in debug mode so I get the print outs that state the alarm is ringing. Would you be able to help? @sdrabenh will also need to test this as it is actually a fix for not passing regression testing.

mathomp4:
> @mathomp4, can you clarify on what was non-zero diff about this? I can certainly test this but will need some help creating a branch analogous to GOCART 2.5-rc1 with this change implemented on top, and instructions on how to build in debug mode so I get the print outs that state the alarm is ringing. Would you be able to help? @sdrabenh will also need to test this as it is actually a fix for not passing regression testing.

@acollow Well, the only reason I call it non-zero-diff is that in my testing it on v12, it was non-zero-diff. But, well, I have not tested it the same way in `develop`. I'll do that.

It's possible/probable that this only changes answers when the GOCART DT is different than heartbeat. That is true in v12 but maybe not in v11 ... Time to test!

PR #350, opened 2025-08-06
Porting GOCART2G to MAPL3 (exercising only one child SS2G)

pchakraborty:


pchakraborty:
Closed in favor of #356

PR #360, opened 2025-09-19
Filter out data gridcomps

pchakraborty:
Several changes in GOCART2G

- Use `MAPL_GridCompGetChildName` and call its run phase if it's not a data gridcomp
- Use macros `_SET/GET_NAMED_PRIVATE_STATE`
- Some more cleanup

### DIFF
--- a/ESMF/GOCART2G_GridComp/GOCART2G_GridCompMod.F90
+++ b/ESMF/GOCART2G_GridComp/GOCART2G_GridCompMod.F90
@@ -11,7 +11,7 @@ module GOCART2G_GridCompMod
    use MAPL_Constants, only: MAPL_GRAV, MAPL_PI
 
    use mapl3g_generic, only: MAPL_GridCompSetEntryPoint, MAPL_GridCompGet, MAPL_GridCompAddSpec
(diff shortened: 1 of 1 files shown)


ESMF/GOCART2G_GridComp/GOCART2G_GridCompMod.F90
```diff
@@ -484,11 +470,16 @@ subroutine Run1 (gc, import, export, clock, RC)
...
 
-      call MAPL_GridCompGet(gc, logger=logger, _RC)
+      call MAPL_GridCompGet(gc, logger=logger, num_children=num_children, _RC)
       call logger%info("Run1: starting...")
-      call MAPL_GridCompRunChildren(gc, phase_name="Run1", _RC)
+      do iter = 1, num_children
+         child_name = MAPL_GridCompGetChildName(gc, iter, _RC)
+         if ((index(child_name, "data")) /= 0) cycle
+         call MAPL_GridCompRunChild(gc, child_name, phase_name="Run1", _RC)
+      end do
```
tclune, line 482: We can think about regexp in MAPL 3.1.

PR #371, opened 2026-01-08
instance RC settings for v12

acollow:


acollow:
Yes, you are correct that @sdrabenh's v12 is out of date. We discussed this at the GOCART and modeling meetings earlier in the week. The plan is to move forward with the initial release of v12 without the GOCART updates. In the meantime, @vbuchard will work on putting out two releases for GOCART 2.5.*. The code will be identical, and include all the necessary updates for v12 but the first release will have the settings for v11 and the second will have the settings for v12. Once all of this is completed, I will test GCM v12 with our GOCART release to determine if additional tuning is necessary, and then subsequently all of the GOCART changes will go into a GCM v12.1 release.

mathomp4:
And I of course meant:

I was just surprised to see both going to `develop`. Sigh. Brain not doing good today.

mathomp4:
> I am closing this PR for the time being and will put in a new one as the pressure lids need to be updated again, and the sea salt and dust emissions factors were lost in the merge to the updated develop.

Probably the safe thing to do! Wait for Bill to finalize and then update :)

PR #381, opened 2026-02-13
v2.5: Fix start-stop and missing GMI_HNO3 in NI AMIP ExtData yaml

mathomp4:
Testing with GOCART v2.5.0 found that it fails to run with AMIP emissions. The culprit was `GMI_HNO3` which was added to `NI2G_GridComp_ExtData.yaml` in https://github.com/GEOS-ESM/GOCART/pull/373 but it wasn't added to `AMIP/NI2G_GridComp_ExtData.yaml`

This PR adds it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,11 +15,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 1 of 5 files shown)


vbuchard (dismissed):
Thanks @mathomp4

vbuchard (approved):
Thanks Matt!

PR #382, opened 2026-02-13
v2.6: Fix start-stop and missing GMI_HNO3 in NI AMIP ExtData yaml

### LINKED ISSUE #383: GOCART v2.5.0: Fails start-stop and Possible UFS issues
## Summary

GOCART v2.5.0 does not pass start-stop regression. GOCART v2.4.0 does. As does v2.4.3. As does v2.5.0-rc.3. As does v2.5.0-rc.4

==> Ahh. v2.5.0-rc.5 fails.

So the current changes to focus on are:

https://github.com/GEOS-ESM/GOCART/compare/v2.5.0-rc.4...v2.5.0-rc.5

## How to Reproduce

1. Clone GEOSgcm v11 `main`
2. Get GOCART v2.5.0 with fix with `mepo checkout bugfix/v2.5-fix-ni-yaml GOCART`
3. Build
4. Set up an experiment (I did c24 AMIP)
5. Go the experiment, set it up as usual with restarts, etc. (I do `makeoneday.bash link allhis`)
6. On an interactive node, go into the `regress/` folder and run `./gcm_regress.j --startstop`

---

## Longwinded Explanation

This is a "high-level" issue.

@tclune reported that our UFS collaborators were seeing non-reproducibility when running with "recent" GOCART (v2.5.0-ish)

So, to test I built GEOSgcm v11 with GOCART v2.5.0. My first run failed because of a missing ExtData entry for AMIP emissions. This is fixed in #381 (and #382 for v2.6.0). Once I fixed this, I found that GOCART v2.5.0 passes layout regression but is *failing* start-stop regression. Because of this, I've drafted the v2.5.1 and v2.6.1 PRs since we need other fixes.

I am now building GEOSgcm v11 with GOCART v2.4.0 to see if that has the same regression failure. I will update that when I see.

---

cc @vbuchard

mathomp4:
This is the v2.6 (aka GEOSgcm v12) equivalent of #381 

---

Testing with GOCART v2.5.0 found that it fails to run with AMIP emissions. The culprit was `GMI_HNO3` which was added to `NI2G_GridComp_ExtData.yaml` in https://github.com/GEOS-ESM/GOCART/pull/373 but it wasn't added to `AMIP/NI2G_GridComp_ExtData.yaml`

This PR adds it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,12 +15,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-## [v2.6.1] - 2026-02-13
+## [v2.6.1] - 2026-02-17
 
 ### Fixed
 
 - Fix missing `GMI_HNO3` entry in `AMIP/NI2G_GridComp_ExtData.yaml` file
-
+- Fix start-stop regression failure in SU GridComp
 
 ## [v2.6.0] - 2026-02-05
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+## [v2.6.1] - 2026-02-13
+
+### Fixed
+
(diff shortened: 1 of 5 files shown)


vbuchard (dismissed):
Thanks @mathomp4

vbuchard (approved):

PR #389, opened 2026-03-03
v12: Set default optics to RRTMGP

mathomp4:
In GEOSgcm v12, we have moved from RRTMG to RRTMGP by default. However, GOCART has continued with RRTMG optics by default.

So, this PR updates it to be RRTMGP by default. This is zero-diff because, at the moment, we do sed shenanigans in `gcm_run.j` to change these lines at run time to RRTMGP.

NOTE: This will require changes in GEOSgcm_App for running with RRTMG. See https://github.com/GEOS-ESM/GEOSgcm_App/pull/878

cc: @sdrabenh

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Update aerosol optics bands files to default to RRTMGP bands, rather than RRTMG (as GEOSgcm v12 has switched to RRTMGP)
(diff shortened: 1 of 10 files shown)


mathomp4:
When this PR merges and GOCART defaults to RRTMGP optics, the corresponding GEOSgcm_App changes are in GEOS-ESM/GEOSgcm_App#878. That PR removes the now-redundant `@RRTMGP_RADIATION` sed block and adds a `@RRTMG_RADIATION` block to downgrade `RRTMGP` → `RRTMG` for L72 runs.

vbuchard (dismissed):

vbuchard (approved):

PR #397, opened 2026-03-21
Feature/two freq extinct 2 3 1

rtodling:
This adds frequencies 470 and 870 for extinctions output for JEDI without modifying the behavor felt by PSAS (fields in gaas-bkg-sfc). This has been put on top of v2.3.1 - I hope we can issue a v2.3.2 with ONLY this change.

**SHOLD NOT GO TO MAIN**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+## [v2.3.2] - 2026-03-23
+
+### Changed
+
+- Updated `GOCART2G_GridComp.rc` to output AOD profile at 470 nm and 870 nm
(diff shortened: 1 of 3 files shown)


rtodling:
Notice this is showing way more changes than needed since it is pointing to main - this should go on top of v2.3.1 for which there will be only a single file change GOCART2G_GridComp.rc

mathomp4:
I've updated this PR to point to our `release/v2.3` branch which is where v2.3.1 came from.

Also, I've made an attempt at a `CHANGELOG.md` entry with the help of Claude, but I'll consult @vbuchard to make sure it's actually correct.

mathomp4:
Also: unsure how to label...as I'm not sure what this change does :)

patricia-nasa (approved):

mathomp4:
@rtodling v2.3.2 has been released:

https://github.com/GEOS-ESM/GOCART/releases/tag/v2.3.2

PR #461, opened 2026-06-22
GOCART2G: update aerosol optics, config, and add regression tests

pchakraborty:
> [!NOTE]  
> This version of GOCART is **not** zero-diff compared to the previous MAPL3 version based on v11, although the differences appear to be very small. The cause for non zero-diffness is still being investigated.

## Summary

Standardize `SetServices` names:
- Revert `DU2G_SetServices`, `SS2G_SetServices`, `GOCART2G_SetServices` → `SetServices` in respective `GridCompMod.F90` files. This will be updated along with regression tests at a later point.

DU2G/SS2G/GOCART2G configs:
- Updated the `yaml` configs to match their `rc` counterparts.

Add regression tests:
- `regression/dust-sa`: standalone dust regression test
- `regression/seasalt-sa`: standalone seasalt regression test
- `regression/gocart+seasalt+dust`: combined GOCART+SS+DU regression test


tclune:
Merging for now at request of @pchakraborty so that we don't lose crucial updates.  Roundoff diffs remain and must be resolved before we merge into develop proper.

PR #477, opened 2026-09-04
implement klid for SU and NI tracers

acollow:
This remedies the fact that radiatively active aerosol tracers are zeroed out above the klid but ammonium, ammonia, and sulfur dioxide are not. This is marked as zero diff as zeros will be introduced for these fields if klid is anything but 0.01 hPa in the instance files. It passes all regression testing.


pcolarco (approved):
Looks good.

PR #478, opened 2026-09-08
Changes to use the split cap.yaml linked to MAPL PR #5371 and Issue #5355

aoloso:
This is to use the split CAP.yaml done in MAPL


pchakraborty (approved):
LGTM

PR #480, opened 2026-09-16
Simplify callback field creation and access

bena-nasa:
In anticipation of MAPL3 we were review how the callback fields were created. After careful analysis it seems the this whole pattern of adding a field via the short name, then accessing via a long name that had to be retrieved from the attribute was completely unnecessary.  We could see no reason for this pattern. The fields added to the callback state were created in `add_aero`. They could simply be given the long name in the first place as the field name and retrieved from state with the long name rather than this indirection.

Now so that this doesn't break the other places outside of go-cart using this, we added a named alias using the short name, and still add the attribute, so that places outside of go-cart can still work as they did. Once this PR is in, we can start cleaning up the places in moist and radiation just like go-cart and retrieve by the long name rather than the short name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Introduce new function to add fields to callback state and reference field when retrieving by long name, attribute logic was unneccessary
+
 ### Fixed
 
 ### Added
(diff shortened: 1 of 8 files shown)


pcolarco:
@bena-nasa I reviewed the code and it looks straightforward what you did. Couple of questions:
- Why do you want to merge with v2.6 of GOCART particularly? Would it be better to merge this with develop? Is there a reason for this capability in older model tags?
- The motivation for this change...I see similar constructs in @GMI. While that is no longer in active development it is still in use. Does this similar change *have* to be implemented there so the code remain functional? Similar for @CARMA, although that is still in active development.

mathomp4:
> @bena-nasa I reviewed the code and it looks straightforward what you did. Couple of questions:
> 
> * Why do you want to merge with v2.6 of GOCART particularly? Would it be better to merge this with develop? Is there a reason for this capability in older model tags?
> * The motivation for this change...I see similar constructs in @gmi. While that is no longer in active development it is still in use. Does this similar change _have_ to be implemented there so the code remain functional? Similar for @carma, although that is still in active development.

@pcolarco I think @bena-nasa  did this to `release/v2.6` only because we weren't sure if/when GOCART v2.7 (aka `develop`) would be released and used in GEOSgcm. Since the GCM still uses v2.6.x The thought was to update there. 

I guess we should also have a similar PR to develop as well

